PR #2041, opened 2019-02-08
HTTP/2 Server Push Issues

MattGill98:
First commit is the main fix. Stream was handling `RST_STREAM` response to push promise before transitioning stream state after sending the push promise, causing `Http2SessionException` because the stream state is incorrectly still `IDLE`. Fixed this by swapping the order of the operations.

Second commit addresses spurious NPE sourced at `org.glassfish.grizzly.http2.Http2SessionOutputSink$OutputQueueRecord.access$100(Http2SessionOutputSink.java:318)`. After splitting the class out I can't reproduce the error. Regardless of fix, this is a good OCD change anyway.

### DIFF
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates and others
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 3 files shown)


modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates and others
```
arjantijms: Shouldn't this keep the "All rights reserved." part, but just moves it to next line?

arjantijms (changes requested):

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
MattGill98: I was under the impression that the `All rights reserved.` section wasn't required as the EPL implies it? Would you be able to show me an example of the format you'd like?

arjantijms: Maybe this one?

https://github.com/eclipse-ee4j/websocket-api/blob/master/api/server/src/main/java/javax/websocket/server/HandshakeRequest.java#L2

MattGill98:
Changes made

arjantijms (approved):

PR #2044, opened 2019-02-12
Removed call to ensureBufferSize that would immediatly be called again

Cousjava:
Removes call to `ensureBufferSize` when the next line calls `wrap`, the first line of which is `ensureBufferSize`.


modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLConnectionContext.java
```diff
@@ -361,7 +361,6 @@ SslResult wrap(final Buffer input, Buffer output,
         
         if (allocator != null && isOverflow) {
             updateBufferSizes();
-            output = ensureBufferSize(output, netBufferSize, allocator);
```
MattGill98, line 364: I believe this is actually required. Should the output buffer be too small, the allocator will be null for the second call (which it shouldn't be going by the assertions).

MattGill98 (changes requested):

PR #2045, opened 2019-02-13
(#1994) Fixing case sensitivity problem with special header content-length

### LINKED ISSUE #1994: Special headers handling in http is case sensitive
when using a `org.glassfish.grizzly.http.HttpResponsePacket.Builder`, the following calls have different behavior:

* `builder.header("Content-Length", 10)`
* `builder.header("content-length", 10)`

The first one sets the builder field `contentLength` to `10` while the second does not as would be expected.

svendiedrichsen:
Fixes #1994 

Signed-off-by: Sven Diedrichsen <sdiedrichsen@yahoo.de>

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpHeader.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpHeader.java
@@ -1147,24 +1147,24 @@ public abstract class HttpHeader extends HttpPacket
             if (mimeHeaders == null) {
                 mimeHeaders = new MimeHeaders();
             }
-            if(!handleSpecialHeaderAdd(name, value)) {
+            if(!handleSpecialHeaderAdd(Header.find(name), value)) {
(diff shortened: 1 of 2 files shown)


smillidge (approved):

modules/http/src/main/java/org/glassfish/grizzly/http/HttpHeader.java
```diff
@@ -1147,8 +1147,9 @@ public final T header(String name, String value) {
             if (mimeHeaders == null) {
                 mimeHeaders = new MimeHeaders();
             }
-            mimeHeaders.addValue(name).setString(value);
-            handleSpecialHeaderAdd(name, value);
+            if(!handleSpecialHeaderAdd(name, value)) {
```
arjantijms: Tiny, tiny remark is there should be a space between `if` and `(`.

arjantijms (approved):

PR #2052, opened 2019-04-09
Builds on both JDK 8 and JDK 11 tests passed on JDK 8.

smillidge:
Tests are failing when ran on JDK 11 but further works needs to be done to fix the JDK 11 test failures. All tests pass on JDK 8.

Signed-off-by: Steve Millidge <steve.millidge@payara.fish>

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/threadpool/DefaultWorkerThread.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/threadpool/DefaultWorkerThread.java
@@ -104,4 +104,8 @@ public class DefaultWorkerThread extends Thread implements WorkerThread {
         this.transactionTimeoutMillis =
                 TimeUnit.MILLISECONDS.convert(timeout, timeunit);
     }
+
+    @Override
+    public void destroy() {
+    }
 }
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -47,7 +47,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.20.1</version>
+                    <version>2.22.1</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
--- a/extras/bundles/grizzly-httpservice-bundle/pom.xml
+++ b/extras/bundles/grizzly-httpservice-bundle/pom.xml
@@ -144,7 +144,7 @@
                         <configuration>
                             <classifier>sources</classifier>
                             <excludeScope>provided</excludeScope>
-                            <excludeArtifactIds>junit</excludeArtifactIds>
+                            <excludeArtifactIds>junit, hamcrest-core</excludeArtifactIds>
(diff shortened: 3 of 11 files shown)


bshannon:
What necessitated adding hamcrest-core to the exclude list?  Did a new version
of a dependency add it?  Shouldn't it only be a test-scoped dependency everywhere?

smillidge:
Hamcrest was being pulled in and compiled into the bundles when they unpack the dependency sources. I added the exclude to prevent that.

modules/grizzly/src/main/java/org/glassfish/grizzly/threadpool/DefaultWorkerThread.java
```diff
@@ -104,4 +104,8 @@ public void setTransactionTimeout(long timeout, TimeUnit timeunit) {
         this.transactionTimeoutMillis =
                 TimeUnit.MILLISECONDS.convert(timeout, timeunit);
     }
+
+    @Override
+    public void destroy() {
```
kofemann, line 109: or remove from the interface, see https://github.com/eclipse-ee4j/grizzly/pull/2055

bshannon: Yes, if there's no calls to destroy() in Grizzly, it would probably be better to
remove it from the interface.  (Or remove any calls, which clearly have no effect.)

bshannon (approved):
Generally fine, but you might want to address the Thread.destroy() comments.

modules/grizzly/src/main/java/org/glassfish/grizzly/threadpool/DefaultWorkerThread.java
smillidge: OK I'll check whether it can be removed from the interface I assume that #2055 passes all tests?

kofemann: there are some test failing, But I guess it's unrelated:

```
INFO: Stopped listener bound to [0.0.0.0:18901]
[ERROR] Tests run: 9, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.745 s <<< FAILURE! - in org.glassfish.grizzly.servlet.HttpServerTest
[ERROR] testStartSecureWithConfiguration(org.glassfish.grizzly.servlet.HttpServerTest)  Time elapsed: 0.433 s  <<< ERROR!
javax.net.ssl.SSLHandshakeException: Remote host closed connection during handshake
	at org.glassfish.grizzly.servlet.HttpServerTest.testStartSecureWithConfiguration(HttpServerTest.java:256)
Caused by: java.io.EOFException: SSL peer shut down incorrectly
	at org.glassfish.grizzly.servlet.HttpServerTest.testStartSecureWithConfiguration(HttpServerTest.java:256)
```

smillidge: If that's JDK 11 there are SSL failures when building on JDK 11. That's the next thing to address.

Issue #1905, opened 2017-04-14
Leaking PooledMemoryManager$PoolHeapBuffer


glassfishrobot:
I'm stumbling upon OutOfMemoryErrors in my app. Checking one of its heap dumps I see there are 31k <tt>byte[]</tt> instances worth of ~190MB and 20k instances of <tt>PooledMemoryManager$PoolHeapBuffer</tt>. When I check with <tt>visualvm</tt> where those <tt>byte[]</tt> come from, I see they are indeed related to Grizzly.

I wonder if this is the same bug as #1804 which was closed because it could not be reproduced.

This is how the top 5 heap offenders look like:

> num #instances #bytes Class description
> --------------------------------------------------------------------------
> 1: 31024 192039272 byte[]
> 2: 66196 5674816 char[]
> 3: 7596 2495312 int[]
> 4: 60430 1450320 java.lang.String
> 5: 19488 1403136 org.glassfish.grizzly.memory.PooledMemoryManager$PoolHeapBuffer
> ...

LEt me know what extra information is needed to investigate this issue.
#### Environment
Ubuntu 16.04

glassfishrobot:
@glassfishrobot Commented 
Reported by dwilches

glassfishrobot:
@glassfishrobot Commented 
dwilches said:
I am using <tt>Jersey 2.25.1</tt> by the way.

glassfishrobot:
@glassfishrobot Commented 
@rlubke said:
I don't think what you're observing is a leak. The PooledMemoryManager, will allocate a percentage (10% by default) of the JVM's heap across a series of buffers of varied sizes. See [1].

In 2.4.0 (soon to be released), This has been reduced to 3%.

You have a couple of choices, both require implementing the DefaultMemoryManagerFactory [2] interface.

1) Create and return a new PooledMemoryManager tuned to your tastes. See [1] for tuning options.
2) Create and return a HeapMemoryManager instance. This is the memory manager that was used as
the runtime default prior to the inclusion of the PooledMemoryManager (it does have slightly worse performance characteristics compared to PooledMemoryManager).

You can then start the JVM with the system property -Dorg.glassfish.grizzly.MEMORY_MANAGER_FACTORY=<your class here>.

Make sure your implementation includes a public no-arg constructor.

[...]

Issue #1805, opened 2015-09-28
Introduce Closeable.closeFuture()


glassfishrobot:
Introduce Closeable.closeFuture() so it will be easier to check/wait the connection close state. Currently it's possible to use CloseListener to check the connection close state, but using CloseListener for Future-like usecase is not very elegant.
#### Affected Versions
[2.3.22]

glassfishrobot:
@glassfishrobot Commented 
Reported by oleksiys

glassfishrobot:
@glassfishrobot Commented 
Was assigned to oleksiys

glassfishrobot:
@glassfishrobot Commented 
@rlubke said:
Changes applied:
master: 78fc5dc
2.3.x: f4661c3

glassfishrobot:
@glassfishrobot Commented 
Marked as **fixed** on Friday, October 16th 2015, 1:14:43 pm

glassfishrobot:
@glassfishrobot Commented 
This issue was imported from java.net JIRA GRIZZLY-1805

glassfishrobot:
* **Issue Imported From:** https://github.com/javaee/grizzly/issues/1805
* **Original Issue Raised By:**@glassfishrobot
* **Original Issue  Assigned To:** @glassfishrobot
* **Original Issue  Closed By:**@glassfishrobot

Issue #1781, opened 2015-06-09
memory leak in java.nio.HeapByteBuffer and org.glassfish.grizzly.memory.HeapMemoryManager$TrimmableHeapBuffer


glassfishrobot:
We are seeing a memory leak in java.nio.HeapByteBuffer and org.glassfish.grizzly.memory.HeapMemoryManager$TrimmableHeapBuffer on issuing concurrent connections to grizzly HttpServer in SSL mode. Any ideas on how to fix/workaround this issue?

We are not seeing a problem in the absence of concurrent connections.
#### Environment
centos, java 7
#### Affected Versions
[2.3.17, 2.3.21]

glassfishrobot:
@glassfishrobot Commented 
Reported by anish-jacob

glassfishrobot:
@glassfishrobot Commented 
balwantbisht said:
We are also seeing this issue in Grizzly 2.3.17 and now updated code to use Grizzly 2.3.21(not tested on production server yet)

Server runs fine for few days and suddenly memory consumption starts growing from 2GB to 10GB within 1-2 hours

The thread dump has some interesting allocation data:

```
num     #instances         #bytes  class name
----------------------------------------------
   1:      67156540     3760766240  org.glassfish.grizzly.memory.HeapMemoryManager$TrimmableHeapBuffer
   2:      67156792     3223526016  java.nio.HeapByteBuffer
   3:          6498      363924016  [I
   4:            91      363247064  [Lorg.glassfish.grizzly.Buffer;
   5:         23319       20458984  [B
   6:        393216        9437184  com.elluminate.util.UnlockedChainHead$Indirect
   7:         50655        7121896  <constMethodKlass>
   8:         50655        6494000  <methodKlass>
   9:         32589        5600000  [C
  10:          5001        5583272  <constantPoolKlass>
...
2038:             1             16  sun.reflect.GeneratedConstructorAccessor50
2039:             1             16  sun.reflect.GeneratedConstructorAccessor65
Total     135162015     7787978552
```

glassfishrobot:
@glassfishrobot Commented 
oleksiys said:
Are you using worker or same thread strategy?
How many threads do you guys have?

glassfishrobot:
@glassfishrobot Commented 
balwantbisht said:
[...]

Issue #1703, opened 2014-08-21
Out Of Memory Issue with 75% of Heap filled by HeapMemoryManager$TrimmableHeapBuffer


glassfishrobot:
I am facing Out Of Memory issues on Glassfish 4.0 (Build 89) server.

Version of grizzly from \glassfish\glassfish\modules\nucleus-grizzly-all.jar is
version="2.3.1"

Heap Analysis shows that about 75-80% of the heap is filled up by:
org/glassfish/grizzly/memory/HeapMemoryManager$TrimmableHeapBuffer (leak suspect)

Whenever I face this issue, there is always at-least one thread with following dump:

"http-listener-3(61)" daemon prio=10 tid=0x00002acbf86c2800 nid=0x47dc runnable [0x00002acc0aae7000]
java.lang.Thread.State: RUNNABLE
at sun.nio.ch.EPollArrayWrapper.epollWait(Native Method)
at sun.nio.ch.EPollArrayWrapper.poll(EPollArrayWrapper.java:269)
at sun.nio.ch.EPollSelectorImpl.doSelect(EPollSelectorImpl.java:79)
at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:87)

*   locked <0x00000000c60349e0> (a sun.nio.ch.Util$2)
*   locked <0x00000000c60349d0> (a java.util.Collections$UnmodifiableSet)
*   locked <0x00000000c60347a8> (a sun.nio.ch.EPollSelectorImpl)
    at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:98)
    at org.glassfish.grizzly.nio.tmpselectors.TemporarySelectorReader.read0(TemporarySelectorReader.java:184)
    at org.glassfish.grizzly.nio.tmpselectors.TemporarySelectorReader.read0(TemporarySelectorReader.java:142)
    at org.glassfish.grizzly.nio.tmpselectors.TemporarySelectorReader.read(TemporarySelectorReader.java:114)
    at org.glassfish.grizzly.nio.tmpselectors.TemporarySelectorReader.read(TemporarySelectorReader.java:76)
    at org.glassfish.grizzly.AbstractReader.read(AbstractReader.java:72)
    at org.glassfish.grizzly.nio.transport.TCPNIOTransportFilter.handleRead(TCPNIOTransportFilter.java:77)
[...]

PR #2053, opened 2019-04-10
Memory leak in DefaultFilterChain, resulting in large amounts of HeapBuffer

Cousjava:
The memory leak appeared to come from the MemoryManager
See issues #1703, #1781, #1805, #1905


bshannon (changes requested):
There's a bunch of changes that just reformat the code, often
resulting in excessively long lines.  Please revert those changes.

As far as I can tell, there's only one change that actually addresses
the problem, at line 483.

arjantijms:
@bshannon the concerns have been addressed by @Cousjava it seems.

I do have a followup question for @Cousjava 

After a quick test it indeed seems to work, but I'm not 100% sure *why* it works. Can you provide some details?

bshannon (approved):

PR #2054, opened 2019-04-10
ssl: introduce SSLEngineConfigurator#setSSLParameters

kofemann:
SSLParameters is a class introduced in java6 which can take over
SSLEngineConfigurator. As a first step, let allow applications to
use SSLParameters to configure SSLEngineConfigurator.

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLEngineConfigurator.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLEngineConfigurator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -23,6 +23,7 @@ import java.util.logging.Logger;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 
+import javax.net.ssl.SSLParameters;
 import org.glassfish.grizzly.Grizzly;
 
 /**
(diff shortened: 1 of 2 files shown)


bshannon (approved):
I suppose this is fine, although it's interesting that SSLParameters supports a lot of
options that are ignored here, and SSLEngine can take an SSLParamaters object
directly, but that isn't done here.  I suspect this code should be updated to handle
additions to SSLEngine in JDK 1.7 and 1.8.

kofemann:
As mentioned in the commit message, this can be the first step. The SSLEngineConfigurator can be updated to build SSLParamaters and then use it when SSLEngineConfigurator#configure method is called. This will make code much simple.

modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLEngineConfigurator.java
```diff
@@ -272,6 +273,21 @@ public SSLEngineConfigurator setWantClientAuth(boolean wantClientAuth) {
         return this;
     }
 
+    /**
+     * Apply {@link SSLParameters} to this SSLEngineConfigurator.
```
arjantijms, line 246: > Apply {@link SSLParameters} to this SSLEngineConfigurator.

Might want to document that only the cipherSuites, protocols, needClientAuth and wantClientAuth properties are applied?

arjantijms (changes requested):

kofemann:
may be make sense to squash this two commits...

mnriem:
@kofemann Have you applied the requested changes?

kofemann:
Are people interested in this changes? In general, looks like grizzly project has very low activity since eclipse took over.

arjantijms:
> Are people interested in this changes? In general, looks like grizzly project has very low activity since eclipse took over.

We definitely have a low amount of resources to work on Grizzly, which is really unfortunate :(

arjantijms (approved):
Let's forget about the requested changes

kofemann:
Ok. Let me rebase and retest. I will update the PR.

mnriem:
@arjantijms  Looks good to me

PR #2055, opened 2019-04-12
threadpool: remove WorkerThread#destroy method

kofemann:
this method was repeating Thread#destroy, which was deprecated
and removed in java11.

Signed-off-by: Tigran Mkrtchyan <tigran.mkrtchyan@desy.de>


bshannon (changes requested):
Need to reconcile this with PR #2052, probably by updating that PR to remove the method,
making this PR unnecessary.

kofemann:
Log time no see... I think this RP still makes sense...

MattGill98 (approved):

PR #2064, opened 2019-05-23
Only fail the build afterwards, allow all non-impacted builds to continue

pzygielo:
I propose such modification of build script to not break build on first failed module (like [in example for current script](https://travis-ci.org/pzygielo/grizzly/builds/536404963#L7288)), but rather to continue with as many modules/tests as possible (like [in example with `-fae` option](https://api.travis-ci.org/v3/job/536402989/log.txt)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: java
+dist: trusty
 jdk:
   - oraclejdk8
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ language: java
 jdk:
   - oraclejdk8
 sudo: false
-script: mvn clean verify
+script: mvn --fail-at-end clean verify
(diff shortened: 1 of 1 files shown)


pzygielo:
May I ask for review, please?

pzygielo:
May I ask for review, please?

MattGill98 (commented):
I think this would be quite helpful! Just one small thing while this line is being changed, should we also run Maven in batch mode with -B?

pzygielo:
> I think this would be quite helpful! Just one small thing while this line is being changed, should we also run Maven in batch mode with -B?

:+1: I think it's good idea to do so.

pzygielo:
> run Maven in batch mode

done

MattGill98 (approved):
Love it!

MattGill98:
Since I can't see how the travis failure would relate to this PR, I'm going to merge it anyway

pzygielo:
> Since I can't see how the travis failure would relate to this PR

Travis was enabled in the state of failing builds indeed.

PR #2068, opened 2019-06-12
Fix max requests count not accurate

Linary:
I modified the parameter maxRequestsCount to 5 and tested it, but found that a connection was closed after processing 6 requests.
So I tracked the code and found that after processing the fifth request, this condition keepAliveContext.requestsProcessed <= maxRequestCountis still true until the sixth request.


Linary:
Anyone review?

MattGill98 (approved):

MattGill98 (changes requested):
On second thoughts I think this will need a header update. Just let me check and I'll get back to you.

MattGill98 (approved):
I think header updates aren't required in individual contributions, so you're all good as far as I'm aware!

PR #2073, opened 2019-08-12
Reenabled Color in Travis Output

pzygielo:
@MattGill98 Apparently I misunderstood [you](https://github.com/eclipse-ee4j/grizzly/pull/2064#pullrequestreview-273615932).
As no interactive task (like `release`) is performed, I assumed it WAS about colors, to save few bytes in log or to not have color-control/escape codes in raw log.

Also, `style.color` property does not seem to not work with batch mode enabled - compare https://travis-ci.org/eclipse-ee4j/grizzly/builds/570837112#L6459 and https://travis-ci.org/pzygielo/grizzly/builds/570988115#L6712.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ dist: trusty
 jdk:
   - oraclejdk8
 sudo: false
-script: mvn --batch-mode -Dstyle.color=always --fail-at-end clean verify
+script: mvn --fail-at-end clean verify


MattGill98:
You appear to be right! Batch mode also removes lines such as the `Progress: xxKb` lines which can take up some space, but it's IMO not worth the lack of color.

This is essentially a revert, so I'm also happy to merge this one quickly ;)

MattGill98 (approved):

pzygielo:
> Batch mode also removes lines such as the `Progress: xxKb` lines which can take up some space, but it's IMO not worth the lack of color.

All those goals could be achieved  with #2074 - i.e. keep colors and have no transfer progres reported :grin:

PR #2085, opened 2020-01-27
Abort filter chain if SSL connection closed

MattGill98:
This will prevent endless loops if the SSL connection is terminated early without enough data.

I struggled to create a programmatic test for this as it's a race condition.


Cousjava (approved):

arjantijms:
@MattGill98 There's some errors here regarding the following:

```
[ERROR] testSimplePendingSSLClientWrites[0](org.glassfish.grizzly.SSLTest)  Time elapsed: 0.002 s  <<< ERROR!
java.net.BindException: Address already in use
	at org.glassfish.grizzly.SSLTest.doTestPendingSSLClientWrites(SSLTest.java:619)
	at org.glassfish.grizzly.SSLTest.testSimplePendingSSLClientWrites(SSLTest.java:221)
```

MattGill98:
The errors are present on master too I believe!

bshannon:
If you can show that the "Address already in use" error is there on master, file a
separate bug for and and then I think we should approve this PR assuming it
solves the original problem.

It ought to be possible to insert a filter in there somewhere that can terminate the
connection early to induce the error.

MattGill98:
@arjantijms is this good to merge now?

MattGill98:
@arjantijms can this be merged??

m0mus (approved):
LGTM

PR #2092, opened 2020-04-07
Fix HTTP/2 memory leaks

mz1999:
During the HTTP/2 stress test, it was found that Http2Stream will not be released after it is closed, causing a large number of Http2Stream objects to accumulate on the Http2Session and eventually running out of memory.

I fixed the problem.


MattGill98 (approved):

MattGill98:
Thanks for the change, and sorry for the late review. If you could address the conflicting changes, I'll endeavour to be a little more on the ball getting this reviewed!

MattGill98:
@mz1999 are you able to resolve the conflicts on this change please?

MattGill98:
Thank you!

PR #2093, opened 2020-04-16
Change to the jakarta.servlet API

stuartwdouglas:


stuartwdouglas:
This will need a version bump as well, I am not sure what the plan is here around versioning.

I needed to delete a couple of tests as there were dependencies on the old API, and also exclude some modules for the same reason.

extras/grizzly-httpservice/pom.xml
```diff
@@ -91,7 +91,7 @@
                             *,
                         </Import-Package>
                         <Export-Package>
-                            javax.servlet;javax.servlet.http;version=2.5,
+                            jakarta.servlet;javax.servlet.http;version=2.5,
```
jGauravGupta, line 94: javax.servlet.http => jakarta.servlet.http

arjantijms:
@stuartwdouglas Grizzly has already been jakarta-ized, so I'm closing this one. Feel free to re-open if there's still things to be merged from here.

PR #2094, opened 2020-04-21
FileCacheFilter stalls

sblasco-nlg:
Cache filter (org.glassfish.grizzly.http.server.FileCacheFilter) stalls when installed on a virtual machine (tested with VMWare and VirtualBox) during some time (idle timeout seconds) and doesn't return any response.

I thought that after FilterChainContext.supend() the resume() method should be called.

So I modified the class this way:

      ctx.suspend();

      output.notifyCanWrite(new WriteHandler() {
        @Override
        public void onWritePossible() throws Exception {
          ctx.resume(ctx.getStopAction());
        }

        @Override
        public void onError(Throwable ex) {
          ctx.resume();
          throw new RuntimeException(ex);
        }
      });

And it works! So I suppose this is the correct way but not sure.


mnriem:
@sblasco @anajosep Is this PR still relevant?

PR #2096, opened 2020-05-27
Clean for deploy

arjantijms:
Cleaned Grizzly project in preparation for 3.0.0-M1 staging.

* Cleaned poms - removed old deploy/release plug-ins overring EE4J parent
* Updated versions in poms
* Formatted poms and java code
* Added Thread.sleep() to tests to limit failures due to races


modules/grizzly/src/test/java/org/glassfish/grizzly/FilterChainReadTest.java
```diff
@@ -16,45 +16,58 @@
...
- * 
+ *
  * @author Alexey Stashok
  */
 @SuppressWarnings("unchecked")
 public class FilterChainReadTest extends TestCase {
-    public static final int PORT = 7785;
+    public static final int PORT = PORT();
+
+    static int PORT() {
```
pzygielo, line 56: When I saw this is left with default visibility I hoped it is reused. Unfortunately I see the same code duplicated several times instead. Why not make initial port base argument of this method (not the best name IMO, as there is already field of that name, and I'd like the method to have a verb) and have only one?

arjantijms: I had the idea of calling the PORT() method from within the code whenever the "cannot bind to port exception occurs". But actually the exception is often not "true", and it's just a matter of a sleep. For now, it mostly needed a quick way to pass the tests and do away with stuck ports.

Better would be to move the port code, AND the connection code, to a test utility class and have all tests use that. That way the get port, sleep, retry and obtain new port, can be kept at one place.

arjantijms: p.s. there's quite a bit more cleaning to do for Grizzly. Specifically the `Connection` type is missing generics everywhere and there's a bunch of JavaDoc warnings when building, as well as forkCount warnings etc.

arjantijms: Also many warnings about deprecated types:

```
[WARNING]
 /home/travis/build/eclipse-ee4j/grizzly/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Stream.java:
[41,29]
 org.glassfish.grizzly.ICloseType in org.glassfish.grizzly has been deprecated
```

MattGill98 (approved):
LGTM

PR #2109, opened 2020-11-06
Fixes for OSGi environment

senivam:
Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>

### DIFF
--- a/modules/bundles/http/pom.xml
+++ b/modules/bundles/http/pom.xml
@@ -70,6 +70,7 @@
                 <configuration>
                     <instructions>
                         <Import-Package>
+                            !android.os,
                             *,
                         </Import-Package>
                         <Export-Package>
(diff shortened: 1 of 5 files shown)


arjantijms (approved):
You beat me to it. LGTM!

senivam:
I had to add -Pstaging to the build configuration (this can be reverted later) due to jakarta.ws.rs (3.0.0). dependency which is still in staging only.

senivam:
actually I had to revert `-Pstaging` back because it appears to be useless - build starts from somewhere else. So, build on Travis will fail untill jakarta.ws.rs is in central. But since release (in Jenkins) is being done using `-Pstaging` profile it could be released before publishing jakarta.ws.rs to central...

arjantijms:
> I had to add -Pstaging to the build configuration (this can be reverted later) due to jakarta.ws.rs (3.0.0). dependency which is still in staging only.

This should already have been pushed out to Maven central.  @spericas can you push the staged artefact out?

arjantijms:
ps jakarta.ws.rs 3.0.0 is on maven central

PR #2112, opened 2020-11-19
Aggregated fixes from Payara

dmatej:
Fixes #2016 and #2111

This is a aggregated PR of Payara's fork of 2.4.4, I hope I didn't skip anything.
Current Payara 5.2020.7-SNAPSHOT with these commits in it's own patch of Grizzly passed following tests:

- all Grizzly's own tests
- all Payara's own tests
- TCK - websockets
- TCK - servlet
- my private test using arquillian and Apache HTTP Client and customized reproducer from #2016 
- HTTP/2 Push reproducer "100 Delphines" (after reconfiguring Payara a bit to use 100 threads for HTTP or more)
  - note - this test causes exceptions with Firefox (remote stream reset), but no exceptions on Chrome

### Merge:

* these changes were based on changes done in #2103 , which should be merged first. I will resolve conflicts after that.

PR #2113, opened 2020-11-20
Updated travis file

dmatej:
The travis file was obsoleted and using Oracle 8u151 would fail AlpnSupportTest in #2112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 language: java
-dist: trusty
-jdk:
-  - oraclejdk8
+os: linux
+dist: focal
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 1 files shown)


dmatej:
Forced push - to avoid dependency issues, becuase master head now depends on Grizzly Bootstrap NPN version which wasn't released yet, so I did several steps back as there is no risk of conflicts between branches.

dmatej:
Ok, this build finally did what I wanted:
- replaced Ubuntu 14.04 with 20.04
- replaced Oracle 8u151 with OpenJDK 8u175

It failed, because it needs changes in #2112 which enhances compatibility with different actual JDK versions.

Issue #1972, opened 2017-09-29
HTTP/2 Upgrade attempt without the HTTP/2 filters in use with a payload could end up timing out.


glassfishrobot:
When attempting an upgrade, content modifiers will be ignored and then an event is fired.  However, when the HTTP/2 filters aren't installed, this event isn't handled.  and thus the logic to handle the upgrade doesn't reset this flag.  Because of these, a request body could be fully buffered, but the last flag on the HttpContent buffer is never set causing any further attempts to read in an attempt to get to EOF, to timeout.

glassfishrobot:
* **Issue Imported From:** https://github.com/javaee/grizzly/issues/1972
* **Original Issue Raised By:**@rlubke
* **Original Issue  Assigned To:** Unassigned

facundovs:
Hi guys, do you have an ETA for the resolution of this issue?

MattGill98:
I'm not sure if it's the same cause, but an EOFException has been fixed in #2001 which might help?

facundovs:
Hi @MattGill98. Thanks for your response but unfortunately, your change seems not to fix this issue.
Until I see, this bug was introduced in the resolution of [this issue](https://github.com/javaee/grizzly/issues/1741). 
Please take a look at [this line](https://github.com/javaee/grizzly/commit/c996ccddcf148210d2ad94992ad6702b3224ab19#diff-52d8b732e03b896f648a8ab59ff33ed3R359) where the code that ignores h2c headers is commented. My suspicion is that this commit was cherrypicked to 2.3.x without taking into account that HTTP2 is supported since 2.4.x.
As consequence, the IncomingHttpUpgradeEvent is triggered but never handled in 2.3.x (because the handling code is just present in 2.4.x) what causes the timeout described above. 
Probably the solution would be to uncomment the h2c header validation in 2.3.x. However, what I am seeing is that in this repository the branch 2.3.x is not present anymore. Was it an omission or will 2.3.x branch no longer be sustained? 

Thanks in advance for your feedback.

MattGill98:
Hi @facundovs,

Your assessment of the issue seems valid. I can't comment on versions myself, although it is possible that 2.3.x is no longer going to be maintained. Sorry I can't be of more help, and I hope this issue gets resolved for you!

scholzi100:
This bug is also present in upcoming version 3.0.0:
[console.log](https://github.com/eclipse-ee4j/grizzly/files/5588039/console.log)
[...]

PR #2119, opened 2020-11-29
Reset ignoreContentModifiers if incoming/outgoing Upgrade does not succeed

### LINKED ISSUE #2117: HTTP Upgrade header causes reading of payload to block, clients timeout if no upgrade filter is available
If a request with a payload and `Upgrade` header is received but the server has no filter to handle this specific upgrade the connection times out. 
This is also reflected in issue #1972, specific to `h2c`. But it does also work with random header values like `123`.

This is not the expected behavior as defined in [RFC 7230](https://tools.ietf.org/html/rfc7230). If the protocol can not be switched:
> A server MAY choose to ignore the order of preference indicated by the
   client and select the new protocol(s) based on other factors, such as
   the nature of the request or the current load on the server.
> [RFC 7230: 6.7 Upgrade](https://tools.ietf.org/html/rfc7230#section-6.7)

> If the server decides to upgrade the connection, it sends back a 101 Switching Protocols response status with an Upgrade header that specifies the protocol(s) being switched to. If it does not (or cannot) upgrade the connection, it ignores the Upgrade header and sends back a regular response (for example, a 200 OK).
> [MDN Web docs: Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#Upgrading_HTTP1.1_Connections)

If no filter for a specific upgrade path is avaliable, it should use the current protocol like nothing happend.

After some testing with and without upgrade filter `org.glassfish.jersey.message.internal.ReaderWriter#readFromAsString(java.io.Reader) ` reads the payload. But if no filter is available some how it stalls while reading from the reader.

scholzi100:
Fixes #2117 and #1972 
>If a request with a payload and Upgrade header is received but the server has no filter to handle this specific upgrade the connection times out.

> `ReaderWriter#readFromAsString(java.io.Reader)` does block because of the underlying Buffer. Its calls to `InputBuffer#read`[*](https://github.com/eclipse-ee4j/grizzly/blob/a2ce7775658e11fbccbb9acd32e2daf2b0799f45/modules/http/src/main/java/org/glassfish/grizzly/http/io/InputBuffer.java#L280) then `InputBuffer#fill()`[*](https://github.com/eclipse-ee4j/grizzly/blob/a2ce7775658e11fbccbb9acd32e2daf2b0799f45/modules/http/src/main/java/org/glassfish/grizzly/http/io/InputBuffer.java#L1058) blocks since `ignoreContentModifiers` was set `true` in 
`onIncomingUpgrade`[*](https://github.com/eclipse-ee4j/grizzly/blob/a2ce7775658e11fbccbb9acd32e2daf2b0799f45/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java#L1346) and was never reset on the upgrade path.

Reset ignoreContentModifiers after upgrade. This may not be a appropriate fix. 

Unit test for the Upgrade case should be added.

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
@@ -1346,12 +1346,16 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
         httpHeader.setIgnoreContentModifiers(true);
 
         ctx.notifyUpstream(HttpEvents.createIncomingUpgradeEvent(httpHeader));
+
+        httpHeader.setIgnoreContentModifiers(false);
     }
 
     protected void onOutgoingUpgrade(final FilterChainContext ctx, final HttpHeader httpHeader) {
         httpHeader.setIgnoreContentModifiers(true);
 
         ctx.notifyUpstream(HttpEvents.createOutgoingUpgradeEvent(httpHeader));
+
+        httpHeader.setIgnoreContentModifiers(false);
     }
 
     protected Buffer encodeHttpPacket(final FilterChainContext ctx, final HttpPacket input) {


mnriem:
@scholzi100  Is this PR still relevant?

PR #2131, opened 2021-08-02
correct the link to eclipse github grizzly page.

anbusampath:


arjantijms:
Change seems fine, but the process requires the ECA to be signed.

anbusampath:
Someone reason I can't use that email address for ESA, I will raise new PR.

PR #2143, opened 2021-10-28
Finer configuration of GZIP Compression

JamesHillyard:
Allows for finer configuration of GZIP Compression by setting the compression level and strategy. 

Tested on Windows 10 JDK 8

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/compression/zip/GZipEncoder.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/compression/zip/GZipEncoder.java
@@ -43,8 +43,8 @@ public class GZipEncoder extends AbstractTransformer<Buffer, Buffer> {
     private static final int TRAILER_SIZE = 8;
 
     private final int bufferSize;
-    private static int compressionLevel;
-    private static int compressionStrategy;
+    private final int compressionLevel;
+    private final int compressionStrategy;
 
     private static final Buffer header;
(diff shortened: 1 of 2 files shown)


modules/grizzly/src/main/java/org/glassfish/grizzly/compression/zip/GZipEncoder.java
```diff
@@ -43,6 +43,8 @@
     private static final int TRAILER_SIZE = 8;
 
     private final int bufferSize;
+    private static int compressionLevel;
```
MattGill98: As these are configured through a constructor, they shouldn't be static

MattGill98 (changes requested):

modules/grizzly/src/main/java/org/glassfish/grizzly/compression/zip/GZipEncoder.java
arjantijms: @JamesHillyard @MattGill98 any chance to get this resolved?

JamesHillyard: Hi @arjantijms, thanks for reminding me about this, I'll aim to get these changes done before the end of the week ready for another review

arjantijms (approved):

MattGill98 (approved):

PR #2151, opened 2021-12-10
Removed deprecated usages of getPeerCertificateChain

dmatej:
A message for the kind reviewer: if you think that it would be good to simply remove those methods now marked as deprecated, I can do that. The  `SSLSession.getPeerCertificateChain` used by the old implementation is not existing in JDK17 any more.

In regard to backward compatibility - calling deprecated methods should be still possible, but implementing the interface means implementing the new version.

Locally all tests passed on JDK11 and JDK17.

### DIFF
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -212,7 +212,7 @@
                         <configuration>
                             <rules>
                                 <requireJavaVersion>
-                                    <version>[1.8,)</version>
+                                    <version>[11,)</version>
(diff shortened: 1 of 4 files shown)


arjantijms (approved):

arjantijms:
>  if you think that it would be good to simply remove those methods now marked as deprecated, I can do that.

Let's leave them in for 1 version, so people who find those methods have disappeared can look at the previous version to see what they moved to.

PR #2159, opened 2022-03-23
Setup GH Workflow for PRs

pzygielo:
I propose for consideration GitHub workflow to run for Pulls, as I can't find job with similar function on ci.eclipse.org.

There are two jobs defined:
1. Perform build without running tests.
2. If above completes successfully - build with tests this time.

I decided to split (temporarily) the build to two jobs, as I find many tests failing from time to time. The problem might be with the tests, might be with the grizzly itself.
Thus the failing job will be left to the discretion of reviewer to decide/ignore.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,43 @@
+#
+# Copyright (c) 2022 Contributors to the Eclipse Foundation
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0,
+# or the Eclipse Distribution License v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+
+name: Eclipse Grizzly NIO CI
+
+on:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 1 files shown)


pzygielo:
May I ask for review, please?

.github/workflows/maven.yml
```diff
@@ -0,0 +1,72 @@
...
+      with:
+        distribution: 'temurin'
+        java-version: ${{ matrix.java_version }}
+    - name: Maven Build
+      run: |
+        mvn --show-version \
+            --no-transfer-progress \
+            --activate-profiles staging \
+            --define skipTests=true \
+            install
```
kofemann, line 41: instead of skipping the  tests you can use `compile` target

kofemann (approved):
LGTM

.github/workflows/maven.yml
pzygielo: But `compile` phase will not compile tests, will it?

This job currently checks that everything is installable (almost-deployable), with the _temporary_ exception for unit tests execution. That's how I see it.

kofemann: That's right. I was mot thinking in direction compile -> compile test + run tests -> package -> verify (or install).  I typically run 'package', and then with a differen step sigh and publish  artifacts.

Anyway, good enough for the kick-off

arjantijms (approved):

PR #2162, opened 2022-03-25
Fixed NPE in toString

dmatej:
- getMessage can return also null char[] array; String.valueOf(char[]) doesn't expect that.

Signed-off-by: David Matějček <dmatej@seznam.cz>

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -254,7 +255,7 @@ public class FilterChainContext implements AttributeStorage {
 
     /**
      * Get the current processing task state.
-     * 
+     *
      * @return the current processing task state.
      */
     public State state() {
@@ -433,7 +434,7 @@ public class FilterChainContext implements AttributeStorage {
 
     /**
      * Get the general Grizzly {@link Context} this filter context wraps.
-     * 
+     *
      * @return the general Grizzly {@link Context} this filter context wraps.
      */
     public final Context getInternalContext() {
@@ -926,7 +927,8 @@ public class FilterChainContext implements AttributeStorage {
         sb.append("connection=").append(getConnection());
         sb.append(", closeable=").append(getCloseable());
         sb.append(", operation=").append(getOperation());
-        sb.append(", message=").append(String.valueOf(getMessage()));
+        // message can be null, then valueOf(char[]) would be executed -> NPE
+        sb.append(", message=").append(String.valueOf(Object.class.cast(getMessage())));
         sb.append(", address=").append(getAddress());
         sb.append(']');
 
@@ -950,7 +952,6 @@ public class FilterChainContext implements AttributeStorage {
         }
     }
 
-    @SuppressWarnings("deprecation")
     public static final class TransportContext {
         private boolean isBlocking;
         CompletionHandler completionHandler;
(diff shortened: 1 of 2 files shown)


dmatej:
Don't merge yet, neither previous nor this version works - it is a trap with generics :-/

kofemann:
Instead of 'don't merge' warning you should change the PR type to 'Draft', then the merge will be disabled until you release it.

pzygielo:
> Don't merge yet, neither previous nor this version works - it is a trap with generics :-/

Perhaps UT would be helpful.
For example of
```java
package org.glassfish.grizzly.filterchain;

import org.junit.Test;

public class FilterChainContextTest {
  @Test
  public void testToStringWithNonNullMessage() {
    var ctx = new FilterChainContext();
    char [] msg = "Grizzly Test".toCharArray();
    ctx.setMessage(msg);

    var str = ctx.toString();

    assert "FilterChainContext [connection=null, closeable=null, operation=NONE, message=Grizzly Test, address=null]".equals(str) : str;
  }

  @Test
  public void testToStringWithNullMessage() {
    var ctx = new FilterChainContext();
    char [] msg = null;
    ctx.setMessage(msg);

    var str = ctx.toString();

    assert "FilterChainContext [connection=null, closeable=null, operation=NONE, message=null, address=null]".equals(str) : str;
  }

  @Test
  public void testToStringWithNullObjectMessage() {
    var ctx = new FilterChainContext();
    Object msg = null;
    ctx.setMessage(msg);

    var str = ctx.toString();

    assert "FilterChainContext [connection=null, closeable=null, operation=NONE, message=null, address=null]".equals(str) : str;
  }
}
```
two tests fail with NPEx on master, and don't on this PR.
Could this trap be shown with the test?

modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
```diff
@@ -926,7 +927,8 @@ public String toString() {
         sb.append("connection=").append(getConnection());
         sb.append(", closeable=").append(getCloseable());
         sb.append(", operation=").append(getOperation());
-        sb.append(", message=").append(String.valueOf(getMessage()));
+        // message can be null, then valueOf(char[]) would be executed -> NPE
+        sb.append(", message=").append(String.valueOf(Object.class.cast(getMessage())));
```
pzygielo, line 931: This produces something like `[C@78e94dcf` for non null message :disappointed:

dmatej: Which is probably correct as we don't want to see gigabyte in logs. Also usually the message is a Buffer instance, not char[]. char[] is probably never used as the message. To me it seems that generics were used wrong here as it can be anything but not anything is usable.

dmatej: The char[] is used as the most concrete valueOf method variant compatible with StringBuilder.append.
Such test would be probably misleading, but yeah, I can do some.

dmatej:
> Instead of 'don't merge' warning you should change the PR type to 'Draft', then the merge will be disabled until you release it.

If I could I would do that, but there is no such option (Drafts disabled on this repo?)

The bug was visible in glassfish tests with full logging and also from Eclipse editor. Simply null typed as <T> means "use the most concrete method compatible, which is with char[] argument, not Object."
Quite interesting, but String.valueOf(char[]) doesn't like null.

dmatej:
Original version vs. trivial test:
image

dmatej:
Note - grizzly-framework tests passed on my machine, seems like some conflict on ports. Also my test was executed after failing tests.

```
Error:  Tests run: 8, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 1.209 s <<< FAILURE! - in org.glassfish.grizzly.TCPNIOTransportTest
Error:  org.glassfish.grizzly.TCPNIOTransportTest.testClose  Time elapsed: 0.237 s
Error:  org.glassfish.grizzly.TCPNIOTransportTest.testThreadInterruptionDuringAcceptDoesNotMakeServerDeaf  Time elapsed: 0.554 s
Error:  org.glassfish.grizzly.TCPNIOTransportTest.testMultiBind  Time elapsed: 0.008 s  <<< ERROR!
java.net.BindException: Address already in use
	at org.glassfish.grizzly.TCPNIOTransportTest.testMultiBind(TCPNIOTransportTest.java:139)
```

arjantijms (approved):

PR #2166, opened 2022-04-12
Attempt to stabilize TCPNIOTransportTest

dmatej:
- closing connections is asynchronous, test did not respect that
- testThreadInterruptionElsewhereDoesNotMakeServerDeaf had two paths to succeed
- using port over 10000 (should ba safer than 9981)


pzygielo:
I haven't read the change yet, but I :heart::
- the result with GH Actions
- locally `TCPNIOTransportTest.testMultiBind` passes too!

arjantijms (approved):

dmatej:
Good - it always passed even before, on my machine, but ...
- on GH failed just one tests, other using the same port passed, so it wasn't anything like "something on build server uses the same port"
- build servers are usually quite slower than my machine, so when anything works with ports/files I am always searching some close->open trap. 

NIO is heavily multithreaded, so it just sends signals like "please stop". Then it stops accepting requests, probably sends some good bye packets, closes input and output ... and here it finally calls an on-close listener if provided - but that doesn't block caller.

PR #2179, opened 2022-12-07
src: drop NullaryFunction in favor or java.util.Supplier

kofemann:
The NullaryFunction has an equivalent in standard java API. So, lets use it.


mnriem:
@kofemann Is this change backwards compatible?

arjantijms:
@mnriem maybe we can just build it, and test it? E.g. by adding it to GlassFish and running all tests that we have?

kofemann:
@mnriem I am pretty sure this is not a drop-in replacement. However, I have no idea which external applications use grizzly internals. So, I agree with @arjantijms, we just have to build packages, document the change and let depending applications to test.

arjantijms (approved):

modules/grizzly/src/main/java/org/glassfish/grizzly/attributes/AttributeBuilder.java
```diff
@@ -61,29 +63,18 @@ public interface AttributeBuilder {
...
-    /**
-     * Create Attribute with name and initializer, which will be called, if Attribute's value is null on a AttributedObject
-     *
-     * @param <T> Type of attribute value
-     * @param name attribute name
-     * @param initializer NullaryFunction, which will be called, if Attribute's value is null on a AttributedObject
-     *
-     * @return Attribute<T>
-     */
-    <T> Attribute<T> createAttribute(String name, org.glassfish.grizzly.utils.NullaryFunction<T> initializer);
```
pzygielo, line 73: With this removed, after switching client code to Supplier, the method left below is used, which is still marked as deprecated. This causes:
```
[WARNING] .../mq/main/mq-broker/broker-core/src/main/java/com/sun/messaging/jmq/jmsserver/service/imq/grizzly/GrizzlyMQPacketFilter.java:[42,13] <T>createAttribute(java.lang.String,org.glassfish.grizzly.attributes.NullaryFunction<T>) in org.glassfish.grizzly.attributes.AttributeBuilder has been deprecated
[INFO] 1 warning
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] .../mq/main/mq-broker/broker-core/src/main/java/com/sun/messaging/jmq/jmsserver/service/imq/grizzly/GrizzlyMQPacketFilter.java: warnings found and -Werror specified
[INFO] 1 error
```

kofemann: Looks like the `depreciation` left in place unintendedly

PR #2183, opened 2023-04-18
Replace deprecated SSLSession.getPeerCertificateChain with SSLSession.getPeerCertificates

eirbjo:
Efforts are underway in the OpenJDK project to remove the long deprecated-for-removal classes in the package `javax.security.cert`. These classes were introduced for backwards compatibility concerns with the unbundled JSSE release for JDK 1.2/1.3, but their use have been discouraged since they were introduced.

It would be good to update Grizzly 2.4.x to not depend on / use these archaic APIs.

See https://bugs.openjdk.org/browse/JDK-8227024 and the corresponding CSR https://bugs.openjdk.org/browse/JDK-8227395

The `SSLSupportImplClass` currently uses `getPeerCertificateChain`, then converts the results to `java.security.cert.X509Certificate`. This seems a bit strange, since the methods could just call `getPeerCertificates` directly, which returns `java.security.cert.X509Certificate[]`.

Merging this PR would allow Grizzly 2.4.x to be used on a future JDK where the deprecated for-removal APIs in `javax.security.cert` are removed.

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLSupportImpl.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/ssl/SSLSupportImpl.java
@@ -113,7 +113,6 @@ public class SSLSupportImpl implements SSLSupport {
             return null;
         }
 
-        // Convert JSSE's certificate format to the ones we need
         Certificate[] jsseCerts = null;
         try {
             jsseCerts = session.getPeerCertificates();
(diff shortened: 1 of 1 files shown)


kofemann:
Looks like too many unrelated changes.

eirbjo:
> Looks like too many unrelated changes.

Oops, something is weird here with the PR, this was not intended. I'll take a look.

eirbjo:
> Oops, something is weird here with the PR, this was not intended. I'll take a look.

Alright, the PR had the wrong base branch. Retargeted to `2.4.4-RELEASE` which was the intended branch.

mnriem:
@eirbjo Is this PR still relevant?

eirbjo:
> @eirbjo Is this PR still relevant?

I would think so. OpenJDK would still like to remove the mentioned classes, and uses of `SSLSession.getPeerCertificateChain` should be updated to use `SSLSession.getPeerCertificates` instead in preparation for this change.

mnriem:
@arjantijms Looks good to me!

arjantijms (approved):

Issue #2215, opened 2024-10-30
Incorrect scheme specified within Origin header during WebSocket Handshake


cgdrake:
Establishing a non-SSL WebSocket connection results in an Origin header being sent by the client which erroneously specifies the scheme as `wss`. In addition, a call to `Handshake.isSecure()` reports true when in fact the connection is insecure.

The following test demonstrates the issue when run again all current versions of Grizzly:
```
package org.glassfish.grizzly.websockets;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertFalse;
import static org.junit.Assert.assertTrue;

import java.net.URI;
import java.net.URISyntaxException;
import java.util.logging.Logger;

import org.glassfish.grizzly.websockets.rfc6455.RFC6455HandShake;
import org.junit.Test;

public class HandShakeTest {
    private static final Logger LOGGER = Logger.getLogger("HandShakeTest");
    private static String SSL = "wss://localhost:8443";
    private static String NON_SSL = "ws://localhost:8080";
    private static String RESOURCE_PATH = "/websocket";

    @Test
    public void testOrigin() throws URISyntaxException {
        // non-ssl
        HandShake handshake = new RFC6455HandShake(new URI(NON_SSL + RESOURCE_PATH));
        LOGGER.info("Handshake: isSecure=" + handshake.isSecure() + ", headers: " + handshake.composeHeaders().getHttpHeader());
        assertEquals(handshake.getOrigin(), NON_SSL);
        assertFalse(handshake.isSecure());
        assertEquals(handshake.getLocation(), NON_SSL + RESOURCE_PATH);

        // ssl
        handshake = new RFC6455HandShake(new URI(SSL + RESOURCE_PATH));
        LOGGER.info("Handshake: isSecure=" + handshake.isSecure() + ", headers: " + handshake.composeHeaders().getHttpHeader());
        assertEquals(handshake.getOrigin(), SSL);
        assertTrue(handshake.isSecure());
        assertEquals(handshake.getLocation(), SSL + RESOURCE_PATH);


    }
}
```

Result:
```
Oct. 30, 2024 4:02:13 P.M. org.glassfish.grizzly.websockets.HandShakeTest testOrigin
INFO: Handshake: isSecure=false, headers: HttpRequestPacket (
   method=GET
   url=/websocket
   query=null
   protocol=HTTP/1.1
   content-length=-1
   headers=[
      Connection=Upgrade
      Host=localhost:8080
[...]

PR #2216, opened 2024-10-30
Issue #2215 Fix invalid Origin header sent by client for non-SSL WebSocket connections

cgdrake:
Fix for incorrect scheme detection within Handshake constructor. See issue #2215.


carryel (commented):
Changes look good.

There are two things that would be good to reflect.
1. It seems that the license header needs to be revised and added. It seems that you can reflect the year by referencing other files and pasting them.
2. I wonder if it is possible to PR to the main branch rather than the master branch.

cgdrake:
@carryel I have corrected the PR to target the main branch and also updated the copyrights.

carryel (approved):

Issue #2217, opened 2024-12-15
Enhances line terminator for chunk-size in Chunked Transfer Coding


carryel:
The chunk size and chunk extension are terminated by a CRLF sequence.

[RFC 9112 Chunked Transfer Coding](https://www.rfc-editor.org/rfc/rfc9112#name-chunked-transfer-coding)

```
  chunk          = chunk-size [ chunk-ext ] CRLF
```

However, Grizzly allows even if the line terminator is only a single LF.

This can lead to unintended behavior with incorrect chunk-ext in reverse proxy solution environments.

Therefore, we should safely provide the ability to only allow CRLF.

carryel:
For convenience, I will leave some of the contents of the commit log.

```
+ When the org.glassfish.grizzly.http.STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112 option is enabled, only CRLF is allowed as the chunk-size line terminator in Chunked Transfer Coding.

Note) The features related to this issue #2217 only work when the STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112 option is enabled, and the existing code base behavior is maintained when options are not enabled.

ex) -Dorg.glassfish.grizzly.http.STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112=true
```

PR #2218, opened 2024-12-16
Issue #2217 "Enhances line terminator for chunk-size in Chunked Transfer Coding

carryel:
+ When the org.glassfish.grizzly.http.STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112 option is enabled, only CRLF is allowed as the chunk-size line terminator in Chunked Transfer Coding.
+ Added testcase depending on whether option is present or not.

`parsingState.checkpoint2`(extra parsing state field) was not used in chunked transfer coding. 
This was added to verify the CR before LF.
If it were not provided as an option, the code logic could be cleaner, but I provided it as an option within the range that does not affect stability and existing operation as much as possible.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        java_version: [ 11 ]
+        java_version: [ 17 ]
 
     steps:
     - name: Checkout for build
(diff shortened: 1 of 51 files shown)


pzygielo:
Unclear target branch - should it be `master` (as is) or `main`?

carryel:
@pzygielo When I look at the source tree, it looks like main is correct. I'm wondering if I should clean up the PR based on main again, or if I should merge it to master first and then move it to main.

carryel:
Moved to https://github.com/eclipse-ee4j/grizzly/pull/2220

PR #2220, opened 2024-12-24
Feature/strict crlf for chunked transfer coding

carryel:
https://github.com/eclipse-ee4j/grizzly/pull/2218 changed to be based on the main branch.
This is a PR for Issue #2217 based on main branch.

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/ChunkedTransferEncoding.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/ChunkedTransferEncoding.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2010, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -34,6 +35,9 @@ import org.glassfish.grizzly.memory.CompositeBuffer;
 import org.glassfish.grizzly.memory.CompositeBuffer.DisposeOrder;
 import org.glassfish.grizzly.memory.MemoryManager;
 
+import java.util.Objects;
+import java.util.Properties;
+
 /**
  * Chunked transfer encoding implementation.
  *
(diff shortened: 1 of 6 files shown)


carryel:
Maven version should be upgraged in Github Actions env.

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.4.1:enforce (enforce-maven) on project grizzly-project: 
Error:  Rule 1: org.apache.maven.enforcer.rules.version.RequireMavenVersion failed with message:
Error:  Detected Maven Version: 3.8.8 is not in the allowed range [3.8.9,).
```

carryel:
@arjantijms @dmatej If possible, please review this as well. This is also a patch for security enhancement.

modules/http/src/main/java/org/glassfish/grizzly/http/ChunkedTransferEncoding.java
```diff
@@ -49,6 +49,9 @@ public final class ChunkedTransferEncoding implements TransferEncoding {
     private static final byte[] LAST_CHUNK_CRLF_BYTES = "0\r\n".getBytes(ASCII_CHARSET);
     private static final int[] DEC = HexUtils.getDecBytes();
 
+    public static final String STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112 = "org.glassfish.grizzly.http.STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112";
+    private static final boolean isStrictChunkedTransferCodingLineTerminatorSet = Boolean.parseBoolean(System.getProperty(STRICT_CHUNKED_TRANSFER_CODING_LINE_TERMINATOR_RFC_9112));
```
OndroMih: Would you please move this initialization from system property to the constructor and change the static field to instance field?

With that, it's more deterministic when the system property is read and the system property can be reliably set from code before creating the instance, not only using a JVM option.

modules/http/src/main/java/org/glassfish/grizzly/http/ChunkedTransferEncoding.java
```diff
@@ -49,6 +49,9 @@ public final class ChunkedTransferEncoding implements TransferEncoding {
...
+
     private final int maxHeadersSize;
 
     public ChunkedTransferEncoding(final int maxHeadersSize) {
```
OndroMih, line 61: If you can, would you create an overloaded constructor that accepts `Properties` and attempts to read the config from there before it reads the config from System properties?

And to complement it, would you also create overloaded constructors with a `Properties` argument in `HttpServerFilter`, `HttpCodecFilter`, and `HttpClientFilter` that pass it to ChunkedTransferEncoding?

If not, I'll try to do that in another PR, but it would really help. With this, it would be then possible to configure this option in constructor, individually for each server config, rather than using a system property for all listeners. Later we would integrate it in GlassFish to allow setting this property via standard GlassFish config for each listenerrather than via a global system property.

carryel: I understand. I'll look into it later.

carryel: @OndroMih I applied your suggestion.

While modifying, I thought a little about the priority of `Properties` and system properties and how values ​​are applied. The current implementation is to apply only one of the two, and to give priority to the `Properties` setting.

PS) Another way is to merge or override the settings.

OndroMih (approved):

Issue #2222, opened 2025-01-31
100-Continue is not correctly responded for chunked requests


hs536:
Grizzly implements handling for the `Expect: 100-Continue` header; however, this handling fails for chunked requests. While a 100-Continue response should be sent upon completion of parsing the header chunk containing `Expect: 100-Continue`, Grizzly does not reach the following handler method and delays the response until the first byte of the subsequent body is received.

https://github.com/eclipse-ee4j/grizzly/blob/c3dcb70416c8f12aac70ff2abdab72a84a1ee5e0/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/HttpHandler.java#L354-L364

This behavior can cause timeouts, particularly when requests are proxied through Apache 2.4.x. Apache 2.4.x, when a request includes the `Expect: 100-Continue` header, waits for a 100-Continue response from the backend server before sending the request body. This creates a deadlock between Grizzly and Apache, resulting in request timeouts.

While this behavior can be worked around by setting `Proxy100Continue Off` in Apache, this prevents the client from verifying server acceptance of the data before transmission.

carryel:
I have a question to make it clearer.

Does this issue refer to chunked transfer coding? For example, I wonder if it refers to a request that includes a `Transfer-Encoding: chunked` header.

When I reviewed Grizzly source codes related to 100-continue support,

- A request that includes a Content-Length header, not chunked transfer coding
- However, if the content body has not been transmitted yet

If the above is satisfied, it is expected to work well.

If it is chunked transfer coding, it does not seem easy to support this feature. 
Also, when I looked at the spec document, the scenario of using chunked transfer coding and `Expect: 100-continue` together seemed a bit unclear.

If you have any reference materials or scenarios, I will think about it further.

hs536:
> Does this issue refer to chunked transfer coding? For example, I wonder if it refers to a request that includes a Transfer-Encoding: chunked header.

Yes. I've uploaded the simple server application and more detailed reproduction instructions below.
https://github.com/hs536/grizzly-2222

carryel:
With the information you provided, I was able to reproduce it successfully.

As expected, when there is content-length, 100-continue works well,
and in the case of chunked transfer coding, I found that it waits for the length to be included in the body before processing 100-continue.

[...]

PR #2224, opened 2025-02-10
Issue #2222 "100-Continue is not correctly responded for chunked requ…

carryel:
…ests" (https://github.com/eclipse-ee4j/grizzly/issues/2222)

Chunked Transfer Coding supports 100-continue.
+ In case of chunked transfer coding, "Expect: 100-continue" is processed immediately after HttpHeader is parsed.
+ Added tests for chunked transfer encoding to HttpContinueTest.
+ All tests passed.

PS)  However, custom sendAcknowledgment() of HttpHandler cannot be supported in that situation.

### DIFF
--- a/modules/http-server/src/test/java/org/glassfish/grizzly/http/server/HttpContinueTest.java
+++ b/modules/http-server/src/test/java/org/glassfish/grizzly/http/server/HttpContinueTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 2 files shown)


carryel:
https://datatracker.ietf.org/doc/html/rfc7231#section-5.1.1 explains it as follows:
```
 A 100-continue expectation informs recipients that the client is
   about to send a (presumably large) message body in this request and
   wishes to receive a 100 (Continue) interim response if the
   request-line and header fields are not sufficient to cause an
   immediate success, redirect, or error response.
```
Examples include:
```
     PUT /somewhere/fun HTTP/1.1
     Host: origin.example.com
     Content-Type: video/h264
     Content-Length: 1234567890987
     Expect: 100-continue
```
Grizzly handles 100-continue well for HTTP headers with Content-Length, but it does not consider cases where the body has chunk-size, such as chunked transfer coding.

I think the chunked transfer coding spec for 100-continue is a little unclear.
For chunked transfer coding, Grizzly tries to parse up to the chunk-size in the body and passes the control to the HttpHandler. However, in the case of 100-continue, it is determined when only the header is transmitted without the body.

Therefore, in the case of chunked transfer coding, we added the ability to process 100-continue immediately after header parsing is finished.
In other cases, after header parsing, control is passed to HttpHandler as before, and 100-continue is determined therein.
In the case of chunked transfer coding, 100-continue is processed before control is given to HttpHandler, so the existing custom HttpHandler#sendAcknowledgment() in Grizzly cannot be processed. I thought this was the best we could do given the current unclear specs and Grizzly implementation.

hs536 (approved):
LGTM

PR #2229, opened 2025-04-19
Addressed null pointer when constructing a string dataframe

kemalturksonmez:
There is a risk of throwing a `NullPointerException` when calling `toString()` on the `DataFrame`. This can occur due to the `bytes` instance variable not being set. This fix will ensure that the `toString()` call prints out the correct payload for `bytes` or defaults to a reasonable value. This is a small fix to address the current behavior, but I'm sure there are other approaches to preventing this issue. For example, converting `DataFrame` to an immutable object so all values are set during initialization. 

To reproduce this bug, initialize a `DataFrame` using `DataFrame(FrameType type)` or one of the `String` based construtors. Finally, I ran into this bug after enabling logging.

### DIFF
--- a/modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
+++ b/modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
@@ -18,7 +18,6 @@ package org.glassfish.grizzly.websockets;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.Objects;
 
 import org.glassfish.grizzly.utils.Charsets;
 
@@ -33,7 +32,6 @@ public class DataFrame {
         return o instanceof DataFrame;
     }
 
-    private static final byte[] EMPTY_ARRAY = new byte[0];
     private String payload;
     private byte[] bytes;
     private final FrameType type;
--- a/modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
+++ b/modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
@@ -99,7 +99,7 @@ public class DataFrame {
     @Override
     public String toString() {
         return new StringBuilder("DataFrame").append("{").append("last=").append(last).append(", type=").append(type.getClass().getSimpleName())
-                .append(", payload='").append(getTextPayload()).append('\'').append(", bytes=").append(Utils.toString(Objects.requireNonNullElse(getBytes(), EMPTY_ARRAY))).append('}').toString();
+                .append(", payload='").append(getTextPayload()).append('\'').append(", bytes=").append(Utils.toString(getBytes())).append('}').toString();
     }
 
     public boolean isLast() {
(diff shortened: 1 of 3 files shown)


OndroMih (approved):

carryel:
@kemalturksonmez How about adding validation to Utils' toString and toLong? I think it could happen in other calls as well.

kemalturksonmez:
> @kemalturksonmez How about adding validation to Utils' toString and toLong? I think it could happen in other calls as well.

I can add tests as needed. The current fix addresses the issue in `DataFrame`. We could update my fix to make it so that `Utils.toString` and `Utils.toLong` can accept `null` as needed. The latter would be a broader fix but would also expand the accepted argument range for the corresponding methods, which could lead to unexpected behavior down the road.

kemalturksonmez:
@carryel @OndroMih I ended up shifting the null handling to the requested methods in `Utils`. If the defaults aren't acceptable and a `null` return would be preferred, I can make the update as needed.

OndroMih (approved):

modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
carryel, line 1: I'll leave a trivial comment.

In the toString() process, `bytes` changed to getBytes(). I wonder if this was intentional. In the case of getBytes(), it seems that `bytes` are always encoded and reproduced when payload exists. In the case of the current modification, there is an advantage of being able to express `bytes` information when there are no `bytes`, but I think that unintended toString() may cause overhead. In other words, I think it would be better to explicitly obtain and output `bytes` when needed. It would also be good to express the current `bytes` plainly like before the modification.

PS) If there are any modifications, it would be good to update the license year of `Utils.java` and `DataFrame.java`.
ex) `Copyright (c) 2010, 2020` -> `Copyright (c) 2010, 2025`

modules/websockets/src/main/java/org/glassfish/grizzly/websockets/DataFrame.java
```diff
@@ -97,7 +97,7 @@ public void respond(WebSocket socket) {
     @Override
     public String toString() {
         return new StringBuilder("DataFrame").append("{").append("last=").append(last).append(", type=").append(type.getClass().getSimpleName())
-                .append(", payload='").append(getTextPayload()).append('\'').append(", bytes=").append(Utils.toString(bytes)).append('}').toString();
+                .append(", payload='").append(getTextPayload()).append('\'').append(", bytes=").append(Utils.toString(getBytes())).append('}').toString();
```
OndroMih:
```suggestion
                .append(", payload='").append(getTextPayload()).append('\'').append(", bytes=").append(Utils.toString(bytes)).append('}').toString();
```

OndroMih: Since the toString() method now handles null value, I agree it's better to revert this to "bytes" rather than "getBytes()".

Issue #1992, opened 2018-01-17
index out of bounds


glassfishrobot:
![image](https://user-images.githubusercontent.com/972959/35024721-13062cc2-fb7c-11e7-9c8b-83867c7fbab9.png)

glassfishrobot:
* **Issue Imported From:** https://github.com/javaee/grizzly/issues/1992
* **Original Issue Raised By:**@icode
* **Original Issue  Assigned To:** Unassigned

jkutes:
Was this every resolved?
We are experiencing same issue, using "grizzly-framework-2.3.35"

smillidge:
I imagine it will need to a reproducer as there is not much information in this issue

mnriem:
@arjantijms As this one is 5+ years old it is time to just close it out

Issue #1744, opened 2015-02-19
NPE during FilterChain execution


glassfishrobot:
Service with grizzly 2.3.17 has 1k rps and after 1 day of work load 100% CPU of one core 6-cores Xeon. I can't reproduce error in test enviroment.
I use grizzly with framework nuts&bolts ([https://github.com/hhru/nuts-and-bolts](https://github.com/hhru/nuts-and-bolts)) and have exceptions in log:

```
Feb 19, 2015 3:42:19 PM org.glassfish.grizzly.filterchain.DefaultFilterChain execute
WARNING: GRIZZLY0013: Exception during FilterChain execution
java.lang.NullPointerException
	at org.glassfish.grizzly.http.server.HttpServerFilter.afterService(HttpServerFilter.java:380)
	at org.glassfish.grizzly.http.server.HttpServerFilter.handleRead(HttpServerFilter.java:260)
	at org.glassfish.grizzly.filterchain.ExecutorResolver$9.execute(ExecutorResolver.java:119)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeFilter(DefaultFilterChain.java:284)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeChainPart(DefaultFilterChain.java:201)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.execute(DefaultFilterChain.java:133)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.process(DefaultFilterChain.java:112)
	at org.glassfish.grizzly.ProcessorExecutor.execute(ProcessorExecutor.java:77)
	at org.glassfish.grizzly.nio.transport.TCPNIOTransport.fireIOEvent(TCPNIOTransport.java:561)
	at org.glassfish.grizzly.strategies.AbstractIOStrategy.fireIOEvent(AbstractIOStrategy.java:112)
	at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.run0(WorkerThreadIOStrategy.java:117)
	at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.access$100(WorkerThreadIOStrategy.java:56)
	at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy$WorkerThreadRunnable.run(WorkerThreadIOStrategy.java:137)
	at org.glassfish.grizzly.threadpool.AbstractThreadPool$Worker.doWork(AbstractThreadPool.java:565)
[...]

PR #2232, opened 2025-07-19
Prevent overflows

pzygielo:
```
Integer.MIN_VALUE - 1 = Integer.MAX_VALUE = 2147483647
```

Check:
- #1744
- #1992

They were closed without resolution.

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/ExecutorResolver.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/ExecutorResolver.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


pzygielo:
- needs #2233

carryel (approved):

modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/ExecutorResolver.java
```diff
@@ -136,7 +137,8 @@ public final int getNextFilter(FilterChainContext context) {
 
         @Override
         public final int getPreviousFilter(FilterChainContext context) {
-            return context.getFilterIdx() - 1;
+            int filterIndex = context.getFilterIdx();
+            return filterIndex != FilterChainContext.NO_FILTER_INDEX ? filterIndex - 1 : filterIndex;
```
pzygielo: @carryel Thanks for checking

Perhaps it would be better to rewrite without negation:
```java
    return filterIndex == FilterChainContext.NO_FILTER_INDEX ? filterIndex : filterIndex - 1;
```

Originally I tried to implement "Decrement only if it's not NO_FILTER_INDEX already, otherwise return unchanged."
WDYT?

carryel: I agree. Please rewrite it as you said!

carryel (approved):

PR #2221, opened 2024-12-24
Update maven.yml for installing maven

carryel:
+ Added a step for installing maven v3.9.9 by referring to https://github.com/eclipse-ee4j/glassfish/blob/master/.github/workflows/build-windows.yml.

PR #2233, opened 2025-07-19
Use Central to get Maven from

pzygielo:
Only the latest versions are served by https://dlcdn.apache.org/maven/maven-3/.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -34,7 +34,7 @@ jobs:
         java-version: ${{ matrix.java_version }}
     - name: Install Maven
       run: |
-        curl -o ./apache-maven-3.9.9-bin.tar.gz https://dlcdn.apache.org/maven/maven-3/3.9.9/binaries/apache-maven-3.9.9-bin.tar.gz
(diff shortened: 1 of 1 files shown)


pzygielo:
May I ask for review, please?

dmatej (approved):

dmatej:
However I believe that the latest-ubuntu already contains preinstalled latest maven 3, see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md and https://github.com/actions/runner-images

pzygielo:
> However I believe that the latest-ubuntu already contains preinstalled latest maven 3, see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md and https://github.com/actions/runner-images

Maybe. But the versions are fixed there. While the latest maven is 3.9.11 - the runner has 3.9.10 and it will not change (until next ubuntu-x).

- Originally (#2159) `mvn` from the runner was used. It was also the latest maven 3 then. Unfortunately, due to #2209 something like #2221 was necessary. Now it's independent of the runner.

PR #2236, opened 2025-08-05
Contribution to protect against invalid usage of headers

breakponchito:
This is a fix to prevent issues when managing headers on grizzly. Based 
on the [RFC-7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.3) on the section 3.3.3 was mention the following:

 If a message is received with both a Transfer-Encoding and a
 Content-Length header field, the Transfer-Encoding overrides the
 Content-Length.  Such a message might indicate an attempt to
 perform request smuggling (Section 9.5) or response splitting
 (Section 9.4) and ought to be handled as an error.  A sender MUST
 remove the received Content-Length field prior to forwarding such
 a message downstream.

 The fix will filter the message and will return Bad Request if both headers are 
 available:

`printf 'POST /simple/rest/hello/sleep?s=5 HTTP/1.1\r Host: localhost\r Transfer-Encoding: chunked\r Content-Length: 5\r \r 0\r \r GET /simple/rest/hello HTTP/1.1\r Host: localhost\r \r ' | nc localhost 8080`

the result from the call will be the Bad Request error:

image

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpServerFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpServerFilter.java
@@ -588,6 +588,12 @@ public class HttpServerFilter extends HttpCodecFilter {
 
         final MimeHeaders headers = request.getHeaders();
 
+        //here we can add validation to prevent Http Server invalid use of headers
+        if (headers.contains(Header.ContentLength) && headers.contains(Header.TransferEncoding)) {
+            request.getProcessingState().error = true;
+            return;
+        }
+
         DataChunk hostDC = null;
 
         // Check for a full URI (including protocol://host:port/)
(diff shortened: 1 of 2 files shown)


carryel:
Thank you for your PR.
After reviewing the proposal, I have two points to consider.

1. RFC-7230 guides Transfer-Encoding to take precedence when both Transfer-Encoding and Content-Length headers exist simultaneously. Grizzly, in fact, does not consider Content-Length when both are present. This is handled internally to prevent misinterpretation or errors.
2. MimeHeaders' contains() method performs a fairly heavy workload. This change would require a bytes-based comparison of all headers for every request. Since the presence of both headers is already detected during header parsing, this step seems more appropriate.

I believe Grizzly's current behavior should be fine for this situation without any additional modifications. I'm curious if I'm missing something.

pzygielo:
What does it mean: FISH-11629? What does it say? Can the community read it?

https://github.com/search?q=org%3Apayara%20FISH-11629&type=code: 0 results

I'd rather not see this identifier referenced in PR title/commit message.

breakponchito:
@pzygielo I recreated the PR after updating the branch name to remove ticket code from my side, here new PR: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2237

pzygielo:
There was no need to close this one as I said nothing about branch name. PR title edit and force-push would do.

PR #2237, opened 2025-08-06
Contribution to protect against invalid usage of headers

breakponchito:
This is a fix to prevent issues when managing headers on grizzly. Based
on the [RFC-7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.3) on the section 3.3.3 was mention the following:

If a message is received with both a Transfer-Encoding and a
Content-Length header field, the Transfer-Encoding overrides the
Content-Length. Such a message might indicate an attempt to
perform request smuggling (Section 9.5) or response splitting
(Section 9.4) and ought to be handled as an error. A sender MUST
remove the received Content-Length field prior to forwarding such
a message downstream.

The fix will filter the message and will return Bad Request if both headers are
available:

```
printf 'POST /simple/rest/hello/sleep?s=5 HTTP/1.1\r Host: localhost\r Transfer-Encoding: chunked\r Content-Length: 5\r \r 0\r \r GET /simple/rest/hello HTTP/1.1\r Host: localhost\r \r ' | nc localhost 8080`
```

the result from the call will be the Bad Request error:

image

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpServerFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpServerFilter.java
@@ -588,6 +588,12 @@ public class HttpServerFilter extends HttpCodecFilter {
 
         final MimeHeaders headers = request.getHeaders();
 
+        //here we can add validation to prevent Http Server invalid use of headers
+        if (headers.contains(Header.ContentLength) && headers.contains(Header.TransferEncoding)) {
+            request.getProcessingState().error = true;
+            return;
+        }
+
         DataChunk hostDC = null;
 
         // Check for a full URI (including protocol://host:port/)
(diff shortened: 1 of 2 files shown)


pzygielo:
https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2236#issuecomment-3157345361:

@carryel:

Thank you for your PR.
After reviewing the proposal, I have two points to consider.

1. RFC-7230 guides Transfer-Encoding to take precedence when both Transfer-Encoding and Content-Length headers exist simultaneously. Grizzly, in fact, does not consider Content-Length when both are present. This is handled internally to prevent misinterpretation or errors.
2. MimeHeaders' contains() method performs a fairly heavy workload. This change would require a bytes-based comparison of all headers for every request. Since the presence of both headers is already detected during header parsing, this step seems more appropriate.

I believe Grizzly's current behavior should be fine for this situation without any additional modifications. I'm curious if I'm missing something.

carryel:
@breakponchito Please review the information I mentioned above and let me know your opinion. If there are no other opinions, I will close this PR.

breakponchito:
@carryel I will start to double check the case, I was on another stuff but now I will have sometime. Please wait until next week

breakponchito:
@carryel I reviewed and confirmed that grizzly is doing correctly the management of the both headers Transfer-Encoding and Content-Lenght, giving precedence to the Transfer-Encoding in case both are available and discarting Content-Lenght. We can close this ticket we don't need add this contribution

carryel:
@breakponchito Thank you for your confirmation. I will close the PR.

Issue #2228, opened 2025-04-02
Grizzly NPE in DefaultFilterChain, this.serverRequest is null


EkamjotKaur19:
We are experiencing a NullPointerException in Grizzly when we started using Java 17. The issue occurs intermittently. Below is the exception trace:

```
Mar 30, 2025 2:42:47 PM org.glassfish.grizzly.filterchain.DefaultFilterChain execute
WARNING: GRIZZLY0013: Exception during FilterChain execution
java.lang.NullPointerException: Cannot invoke "org.glassfish.grizzly.http.server.Request.getRequestExecutor()" because "this.serverRequest" is null
	at org.glassfish.grizzly.http.server.io.ServerInputBuffer.getThreadPool(ServerInputBuffer.java:91)
	at org.glassfish.grizzly.http.io.InputBuffer.invokeHandlerOnProperThread(InputBuffer.java:1075)
	at org.glassfish.grizzly.http.io.InputBuffer.append(InputBuffer.java:969)
	at org.glassfish.grizzly.http.server.HttpServerFilter.handleRead(HttpServerFilter.java:254)
	at org.glassfish.grizzly.filterchain.ExecutorResolver$9.execute(ExecutorResolver.java:95)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeFilter(DefaultFilterChain.java:260)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeChainPart(DefaultFilterChain.java:177)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.execute(DefaultFilterChain.java:109)
	at org.glassfish.grizzly.filterchain.DefaultFilterChain.process(DefaultFilterChain.java:88)
	at org.glassfish.grizzly.ProcessorExecutor.execute(ProcessorExecutor.java:53)
	at org.glassfish.grizzly.nio.transport.TCPNIOTransport.fireIOEvent(TCPNIOTransport.java:515)
	at org.glassfish.grizzly.strategies.AbstractIOStrategy.fireIOEvent(AbstractIOStrategy.java:89)
	at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.run0(WorkerThreadIOStrategy.java:94)
	at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.access$100(WorkerThreadIOStrategy.java:33)
[...]

Issue #2214, opened 2024-10-21
NPE when closing OutputBuffer


MarcelMerkle:
Hi,

we experienced an issue when trying to resume an asynchronous response in our HTTP server. The issue was caused by an NPE that was thrown when trying to close the `OutputBuffer`:

```
java.lang.NullPointerException: Cannot invoke "org.glassfish.grizzly.http.HttpContext.getOutputSink()" because "this.httpContext" is null
        at org.glassfish.grizzly.http.io.OutputBuffer.blockAfterWriteIfNeeded(OutputBuffer.java:906)
        at org.glassfish.grizzly.http.io.OutputBuffer.close(OutputBuffer.java:684)
        at org.glassfish.grizzly.http.server.NIOOutputStreamImpl.close(NIOOutputStreamImpl.java:76)
        at org.glassfish.jersey.message.internal.CommittingOutputStream.close(CommittingOutputStream.java:251)
        at org.glassfish.jersey.message.internal.OutboundMessageContext.close(OutboundMessageContext.java:565)
        at org.glassfish.jersey.server.ContainerResponse.close(ContainerResponse.java:403)
        at org.glassfish.jersey.server.ServerRuntime$Responder.writeResponse(ServerRuntime.java:763)
        at org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ServerRuntime.java:398)
        at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:388)
        at org.glassfish.jersey.server.ServerRuntime$AsyncResponder$3.run(ServerRuntime.java:926)
        at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
        at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:244)
        at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:266)
        at org.glassfish.jersey.server.ServerRuntime$AsyncResponder.resume(ServerRuntime.java:963)
        at org.glassfish.jersey.server.ServerRuntime$AsyncResponder.resume(ServerRuntime.java:914)
```
We are using grizzly-http-server version 4.0.2.

This is the first time we saw such an issue although we are using this version for several months now.

[...]

PR #2240, opened 2025-08-27
Issue #2228 "Grizzly NPE in DefaultFilterChain, this.serverRequest is…

carryel:
… null" (https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2228)

Issue #2214 "NPE when closing OutputBuffer" (https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2214)

+ Fixed an issue where a new request was unintentionally processed as a suspend request because it was referencing a recycled request.

+ Improvements to the use of recycled InputBuffer and OutputBuffer and concurrent access to them. When a request and response are completed and recycled, the associated ServerInputBuffer and ServerOutputBuffer are also recycled. However, if another thread references the InputBuffer and OutputBuffer, an NPE may occur when using these recycled instances. We consider the possibility of unintended usage after recycling, and the possibility of concurrent calls to recycle and use. I did not modify InputBuffer and OutputBuffer to be completely thread-safe; Improvements have been made only to the areas mentioned in the issues above.

+ Patched an NPE that occurred when calling toString() on a recycled HttpRequestPacket instance (preventing an NPE when obtaining the methodBytes of Method).

+ Patched for TransportFilter's exceptionOccurred() not propagating to filter chain (it seems that there was no apparent intention to prevent propagation).

### DIFF
--- a/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/io/ServerOutputBuffer.java
+++ b/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/io/ServerOutputBuffer.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2012, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -17,6 +18,7 @@
 package org.glassfish.grizzly.http.server.io;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.concurrent.Executor;
 import java.util.logging.Level;
 
@@ -24,16 +26,17 @@ import org.glassfish.grizzly.CompletionHandler;
 import org.glassfish.grizzly.WriteResult;
 import org.glassfish.grizzly.filterchain.FilterChainContext;
 import org.glassfish.grizzly.http.io.OutputBuffer;
+import org.glassfish.grizzly.http.server.Request;
 import org.glassfish.grizzly.http.server.Response;
 import org.glassfish.grizzly.localization.LogMessages;
 
 public class ServerOutputBuffer extends OutputBuffer {
 
-    private Response serverResponse;
+    private volatile Response serverResponse;
 
     public void initialize(final Response response, final FilterChainContext ctx) {
-        super.initialize(response.getResponse(), response.isSendFileEnabled(), ctx);
         this.serverResponse = response;
+        super.initialize(response.getResponse(), response.isSendFileEnabled(), ctx);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


modules/http-server/src/main/java/org/glassfish/grizzly/http/server/io/ServerOutputBuffer.java
```diff
@@ -17,23 +18,25 @@
...
 
 public class ServerOutputBuffer extends OutputBuffer {
 
-    private Response serverResponse;
+    private volatile Response serverResponse;
 
     public void initialize(final Response response, final FilterChainContext ctx) {
-        super.initialize(response.getResponse(), response.isSendFileEnabled(), ctx);
         this.serverResponse = response;
+        super.initialize(response.getResponse(), response.isSendFileEnabled(), ctx);
```
avpinchuk, line 39: Why volatile write happens before `super` initialization?

carryel: @avpinchuk Thanks for reviewing this.
That's right. This order change doesn't currently have any significant implications for ServerOutputBuffer/OutputBuffer. It was simply done to maintain code consistency with ServerInputBuffer/InputBuffer.

https://github.com/carryel/grizzly/blob/439c8ec2eb92c54b1920c4156f44e3ee017aa829/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/io/ServerInputBuffer.java#L41-L43

It is also to prevent mistakes in the future when ServerOutputBuffer's method is called again through super.initialize (like ServerInputBuffer/InputBuffer).

> super.initialize() -> ServerInputBuffer.someMethod()

https://github.com/carryel/grizzly/blob/439c8ec2eb92c54b1920c4156f44e3ee017aa829/modules/http/src/main/java/org/glassfish/grizzly/http/io/InputBuffer.java#L186-L189

https://github.com/carryel/grizzly/blob/439c8ec2eb92c54b1920c4156f44e3ee017aa829/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/io/ServerInputBuffer.java#L81-L85

The reason for changing it to volatile is that, although rare, recycle(), which initializes the variable to null, can be called concurrently from a different thread than the one referencing the variable.

carryel:
This PR is a workaround patch rather than a fundamental fix, and I've confirmed that ignoring recycled requests and processing them as new requests resolves the issue properly. However, I haven't yet discovered the root cause of this issue and am not sure about it.
I think recycle for request and execution for new url should not happen at the same time within the same connection.

Therefore, I believe it would be best to further investigate the root cause before reviewing or applying this PR.
Let me look into it further.

carryel:
I think I found the root cause of the issue.
It appears to be an issue with the attribute holder that stores the request.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/HttpServerFilter.java#L121

For example, after calling httpRequestInProgress.remove(context) and then calling httpRequestInProgress.get(context) to test for deletion, I observed in the log that in certain cases, the attribute value was not properly deleted and remained.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/HttpServerFilter.java#L332

For the AttributeHolder issue, I'll continue the topic in a separate issue.

carryel:
This issue will no longer occur automatically once https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2246 is resolved.
However, the main change is that ServerInputBuffer and ServerOutputBuffer have been improved to be more secure in multithreaded environments.
This is because these two InputBuffer and OutputBuffer can be accessed from multiple threads via NIOInputStream, NIOReader, NIOOutputStream, and NIOWriter.

carryel:
Could you please review this PR when you have time?

While there appear to be a lot of changes, the logic itself remains largely unchanged, with only the following improvements:
- Checking for initialization: While initialization is always performed under normal circumstances, when async reads or writes occur in application code, Grizzly terminates the request and recycles it. However, an NPE can occur in the application that still holds the reference. While this is not a normal situation, we've improved the handling to ensure safer handling.
- Added FINE/FINEST log messages to facilitate future debugging.
- Replicated existing test cases to create an abnormal situation (normally, close() should be called before response.resume() in the application handler, but close() is intentionally called after resume() in testcases). I intentionally triggered an NPE and verified that this exception was handled gracefully.

PS) If possible, I'd like to see these improvements included in a new Grizzly release.

avpinchuk (approved):

Issue #2241, opened 2025-09-12
Grizzly doesn't properly strip whitespace from header values


kenballus:
## The Bug

From RFC 9112:
> A field line value might be preceded and/or followed by optional whitespace (OWS); a single SP preceding the field line value is preferred for consistent readability by humans. The field line value does not include that leading or trailing whitespace: OWS occurring before the first non-whitespace octet of the field line value, or after the last non-whitespace octet of the field line value, is excluded by parsers when extracting the field line value from a field line.

Grizzly does not follow this rule. Grizzly's HTTP parser treats trailing whitespace on a header value as part of the extracted value.

## To reproduce

1. Start a Grizzly server that echoes back received header values, such as [this one](https://github.com/narfindustries/http-garden/).

2. Send it a request with a header with trailing whitespace, and observe that it makes its way into the header value:
```sh
printf 'GET / HTTP/1.1\r\nHost: a\r\nTest: a\t\r\n\r\n' | \
    ncat localhost 80 | \
    grep '"headers"' | \
    jq '.headers[1][1]' | \
    xargs echo | \
    base64 -d | \
    od -tcx1
```
```
0000000   a  \t
         61  09
0000002
```

carryel:
I think I can look into this topic. I'll review it and write back.

carryel:
I've confirmed that only SP is currently considered when parsing header values.

Reviewing the spec below, it appears that HTAB should be also considered.

https://www.rfc-editor.org/rfc/rfc9110.html#whitespace
```
  OWS            = *( SP / HTAB )
                 ; optional whitespace
```

I will fix this soon and submit a PR.

carryel:
@kenballus The patch has been applied. Could you please check with the latest version of the main branch? It worked fine when I tested it.

kenballus:
I have confirmed that the fix works. Thanks!

PR #2242, opened 2025-09-13
Fixed issue #2241 "Grizzly doesn't properly strip whitespace from header values"

carryel:
Trailing HTAB should be removed during the header values ​​parsing process.

https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values
```
...
A field value does not include leading or trailing whitespace. 
...
```

https://www.rfc-editor.org/rfc/rfc9110.html#name-whitespace
```
  OWS            = *( SP / HTAB )
                 ; optional whitespace
```

### DIFF
--- a/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
+++ b/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
@@ -150,6 +150,7 @@ public class HttpRequestParseTest {
         headers.put("Trailing-OWS-Header1", new Pair<>("some-value  ", "some-value"));
         headers.put("Trailing-OWS-Header2", new Pair<>("some-value\t\t", "some-value"));
         headers.put("OWS-Header", new Pair<>(" \t \t some-value \t \t ", "some-value"));
+        headers.put("Mixed-OWS-Header", new Pair<>(" \t \t some- \t \t value \t \t ", "some- \t \t value"));
         doHttpRequestTest("POST", "/index.html", "HTTP/1.1", headers, "\r\n");
     }
 
--- a/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
+++ b/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2010, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
```diff
@@ -140,6 +141,19 @@ public void testSimpleHeadersPreserveCase() throws Exception {
...
+    public void testLeadingOrTrailingWhitespaceFromHeaderContentValues() throws Exception {
+        Map<String, Pair<String, String>> headers = new HashMap<>();
+        headers.put("Host", new Pair<>("localhost", "localhost"));
+        headers.put("Leading-OWS-Header1", new Pair<>("  some-value", "some-value"));
+        headers.put("Leading-OWS-Header2", new Pair<>("\t\tsome-value", "some-value"));
+        headers.put("Trailing-OWS-Header1", new Pair<>("some-value  ", "some-value"));
+        headers.put("Trailing-OWS-Header2", new Pair<>("some-value\t\t", "some-value"));
+        headers.put("OWS-Header", new Pair<>(" \t \t some-value \t \t ", "some-value"));
+        headers.put("Mixed-OWS-Header", new Pair<>(" \t \t some- \t \t value \t \t ", "some- \t \t value"));
+        doHttpRequestTest("POST", "/index.html", "HTTP/1.1", headers, "\r\n");
```
pzygielo, line 154: Unclear for me.

I see it's tested that headers with `WS`es are still parsable.
But is there a test to verify that values are actually stripped of `WS`es?

carryel: I'm not sure I understood your question correctly.
I tried to respect the existing testing approach. I'd like to explain the test logic a bit more.

The Pair object contains a first and second value. The former represents the value sent by the client, and the latter represents the actual parsed result expected by Grizzly.

For example, below is
https://github.com/carryel/grizzly/blob/1adf6be39541721578c547935db3a0d46c0463c6/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java#L151

The client sends `some-value\t\t` and expects the server to interpret it as `some-value` with the trailing ows removed.

The client sends the first value. https://github.com/carryel/grizzly/blob/1adf6be39541721578c547935db3a0d46c0463c6/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java#L435-L446

The server checks the result through the last HTTPRequestCheckFilter filter. It expects the second value, and compares it with the value of the header(that will be sent in the response to the client as is, if it is an echo server).

https://github.com/carryel/grizzly/blob/1adf6be39541721578c547935db3a0d46c0463c6/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java#L413-L414

https://github.com/carryel/grizzly/blob/1adf6be39541721578c547935db3a0d46c0463c6/modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java#L505

pzygielo: Thank you! The `Pair`. I was blind.

pzygielo (approved):

PR #2244, opened 2025-09-17
Fixed for TransportFilter's exceptionOccurred() not propagating to fi…

carryel:
…lter chain (it seems that there was no apparent intention to prevent propagation).

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/TransportFilter.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/TransportFilter.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -179,6 +180,19 @@ public class TransportFilter extends BaseFilter {
         }
 
         return null;
+    }
+
+    /**
+     * Delegates exception operation to {@link Transport}'s specific transport filter.
+     */
+    @Override
+    public void exceptionOccurred(FilterChainContext ctx, Throwable error) {
+
+        final Filter transportFilter0 = getTransportFilter0(ctx.getConnection().getTransport());
+
+        if (transportFilter0 != null) {
+            transportFilter0.exceptionOccurred(ctx, error);
+        }
     }
 
     /**
--- a/modules/grizzly/src/test/java/org/glassfish/grizzly/FilterChainTest.java
+++ b/modules/grizzly/src/test/java/org/glassfish/grizzly/FilterChainTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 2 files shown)


carryel:
I have separated the independent fix and made a PR at https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2240.

carryel:
The Filter interface has the following representative methods:

handleRead, handleWrite, handleConnect, handleAccept, handleEvent, handleClose, exceptionOccurred, ...

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/Filter.java#L133-L141

The implementations of this Filter, TCPNIOTransportFilter, o.g.g.h.HttpServerFilter, and o.g.g.h.s.HttpServerFilter, each handle the above methods. Just as handleRead() is called in a chain when an OP_READ event occurs, if exceptionOccurred() is called in a chain when an exception occurs, post-processing can be performed at each filter location when an exception occurs.

This chained call requires the implementation of the relevant PR. Without this change, an empty function is called, and exceptionOccurred() is not propagated to filters(TCPNIOTransportFilter, UDPNIOTransportFilter), as shown below.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/BaseFilter.java#L109-L110

For reference, an example of how exceptionOccurred() can be raised is:

ctx.fail(...)

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java#L811

When an unexpected exception occurs during filter chain execution

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/227418b2fe2e3b1176d9bbe3ded9a42bfbea8d0e/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/DefaultFilterChain.java#L141

and so on.

carryel:
The exceptionOccurred() method is called from the throwChain() method of DefaultFilterChain:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c6c7d167d8642631629049eaf901eda0fa4cb0aa/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/DefaultFilterChain.java#L419-L427

The throwChain() method is called from the following two places:

One is called when chaining calls are unexpected:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c6c7d167d8642631629049eaf901eda0fa4cb0aa/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/DefaultFilterChain.java#L139-L145

The other is called through the explicit fail() method of FilterChainContext:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c6c7d167d8642631629049eaf901eda0fa4cb0aa/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/DefaultFilterChain.java#L398-L400

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c6c7d167d8642631629049eaf901eda0fa4cb0aa/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java#L811-L813

Unless I missed something in my analysis, even if the fix is ​​applied, it will not cause other issues.

avpinchuk (approved):

PR #2249, opened 2025-10-02
(Trivial) Disable the test for attribute holder's performance testing

carryel:
This can sometimes take several tens of seconds depending on the system environment and performance like this:
https://ci.eclipse.org/glassfish/view/Grizzly/job/Grizzly-master_nightly/2791/


avpinchuk (approved):

pzygielo:
I'd prefer https://github.com/junit-team/junit4/wiki/Ignoring-tests if this is done in code.

Issue #2250, opened 2025-10-02
Replace SimpleDateFormat used with ThreadLocal into java.time packages


carryel:
This issue is derived from https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2206.

Because SimpleDateFormat is not thread-safe, it's often used in conjunction with ThreadLocal.

DateTimeFormatter, introduced in Java 8, is immutable and thread-safe.
With the removal of ThreadLocal, the use of SimpleDateFormat can be migrated to the java.time package.

PR #2251, opened 2025-10-06
Replace SimpleDateFormat used with ThreadLocal into java.time packages #2250

carryel:
java.text.SimpleDateFormat is now deprecated in Grizzly and ThreadLocal, which was used with SimpleDateFormat, has also been removed.

+ Applied to Cookie utilities.
+ Added HttpDateFormat, which replaces FastHttpDateFormat.
+ Added DateTimeFormatters, which replace SimpleDateFormats.
+ Applied to http/http2/http-server modules.
+ Re-enabled ApacheLogFormatTest, which was ignored in the JDK17 build.

### DIFF
--- a/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/accesslog/ApacheLogFormat.java
+++ b/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/accesslog/ApacheLogFormat.java
@@ -775,7 +775,7 @@ public class ApacheLogFormat implements AccessLogFormat {
             }
 
             final DateTimeFormatter format = dateTimeFormatter.withZone(timeZone);
-            return builder.append(format.format(Instant.ofEpochMilli(timeStamp.getTime())));
+            return builder.append(format.format(timeStamp.toInstant()));
         }
 
         @Override
--- a/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/accesslog/ApacheLogFormat.java
+++ b/modules/http-server/src/main/java/org/glassfish/grizzly/http/server/accesslog/ApacheLogFormat.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2014, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -18,7 +19,9 @@ package org.glassfish.grizzly.http.server.accesslog;
 
 import static java.util.logging.Level.WARNING;
 
-import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.Iterator;
@@ -131,7 +134,7 @@ import org.glassfish.grizzly.http.util.MimeHeaders;
  * <ul>
  * <li>When "<code>format</code>" is left unspecified, the default <code>%t</code> format
  * <code>[yyyy/MMM/dd:HH:mm:ss Z]</code> is used</li>
- * <li>When "<code>format</code>" is specified, the given pattern <b>must</b> be a valid {@link SimpleDateFormat}
+ * <li>When "<code>format</code>" is specified, the given pattern <b>must</b> be a valid {@link DateTimeFormatter}
  * pattern.</li>
  * <li>When "<code>@timezone</code>" is left unspecified, the {@linkplain TimeZone#getDefault() default time zone} is
  * used.</li>
@@ -179,7 +182,7 @@ import org.glassfish.grizzly.http.util.MimeHeaders;
 public class ApacheLogFormat implements AccessLogFormat {
 
     /* The UTC time zone */
-    private static final TimeZone UTC = TimeZone.getTimeZone("UTC");
+    private static final ZoneId UTC = ZoneId.of("UTC");
 
     /** A {@link String} representing our version of Apache's <em>common</em> format. */
     public static final String COMMON_FORMAT = "%h - %u %t \"%r\" %s %b";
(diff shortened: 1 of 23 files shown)


modules/http-server/src/main/java/org/glassfish/grizzly/http/server/accesslog/ApacheLogFormat.java
```diff
@@ -757,9 +774,8 @@ StringBuilder format(StringBuilder builder, Request request, Response response,
                 return builder.append('-');
             }
 
-            final SimpleDateFormat format = simpleDateFormat.get();
-            format.setTimeZone(timeZone);
-            return builder.append(format.format(timeStamp));
+            final DateTimeFormatter format = dateTimeFormatter.withZone(timeZone);
+            return builder.append(format.format(Instant.ofEpochMilli(timeStamp.getTime())));
```
avpinchuk: Why not use `timeStamp.toInstant()` instead of `Instant.ofEpochMillis(...)`?

modules/http-server/src/main/java/org/glassfish/grizzly/http/server/util/DateTimeFormatters.java
```diff
@@ -0,0 +1,56 @@
...
+        if (formats != null) {
+            return formats;
+        }
+
+        return new DateTimeFormatters();
+    }
+
+    private final DateTimeFormatter[] f;
+
+    public DateTimeFormatters() {
```
avpinchuk: It's better to make the constructor private.

It is also better to give a more expressive name to the variable `f`, e.g. `formatters`.

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
```diff
@@ -0,0 +1,195 @@
...
+                // should never reach this line
+            }
+        }
+        return cachedStringDate;
+    }
+
+    /**
+     * Get the current date in HTTP format.
+     */
+    public static byte[] getCurrentDateBytes() {
```
avpinchuk, line 109: IMO this method is dangerous.

Please try something similar (if I haven't got anything wrong)

```java
public static byte[] getCurrentDateBytes() {
        final long now = System.currentTimeMillis();
        final long diff = now - nextGeneration;

        if (diff < 0 || (diff < 5000 && isGeneratingNow.get())) {
            return currentDateBytes;
        }

        while (!isGeneratingNow.compareAndSet(false, true)) {
            Thread.onSpinWait();
        }

        if (now > nextGeneration) {
            currentDateBytes = toCheckedByteArray(new StringBuilder(FORMATTER.format(Instant.ofEpochMilli(now))));
            nextGeneration = now + 1000;
        }

        isGeneratingNow.set(false);

        return currentDateBytes;
    }
```

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
```diff
@@ -0,0 +1,195 @@
...
+     * Instant on which the currentDate object was generated.
+     */
+    private static volatile long nextGeneration;
+
+    private static final AtomicBoolean isGeneratingNow = new AtomicBoolean();
+
+    /**
+     * Current formatted date as byte[].
+     */
+    private static volatile byte[] currentDateBytes;
```
avpinchuk: This may be non-volatile

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Current formatted date.
+     */
+    private static volatile String cachedStringDate;
```
avpinchuk: This may be non-volatile too

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
carryel: Your suggested approach is probably correct. I'll look into it a bit more. I know it's risky, but I allowed it because the logic didn't seem to prioritize consistency. However, I think my approach wasn't a good one.

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
carryel: I'd like to ask for clarification. Is this because the `Happens-Before` Guarantee/Relationships affect non-volatile variables?
I missed that.

avpinchuk: > Is this because the `Happens-Before` Guarantee/Relationships affect non-volatile variables?

Yes. Let two variables: non-volatile `a` and volatile `b` and two threads 

```
Thread 1:
a = value1
b = value2 // volatile write

Thread 2:
read b // volatile read
read a
```

Then if `read b` = `value2` then `read a` guaranteed to see `value1`.

avpinchuk: @carryel If you have any questions, I'm always happy to help ;)

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
carryel: I considered several other methods, including ReentrantLock and conditional changes, but your suggestion seemed the most beautiful and efficient, so I adopted it as is.

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
carryel: I resubmitted the PR, removing the unnecessary volatile declaration.

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpDateFormat.java
carryel: I resubmitted the PR, removing the unnecessary volatile declaration.

avpinchuk (approved):

PR #2260, opened 2025-10-29
Fix Test scenarios that were failing in JDK 25

### LINKED ISSUE #2259: Test failures appearing in JDK 25
Test failures appearing in JDK 25 by stricter enforcement of networking protocols and resource management in the recent JDK.

- An unexpected "Premature EOF" (End of File) exception
- javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)

-----
```
java.io.IOException: Premature EOF
	at java.base/sun.net.www.MeteredStream.justRead(MeteredStream.java:58)
	at java.base/sun.net.www.MeteredStream.read(MeteredStream.java:135)
	at java.base/java.io.FilterInputStream.read(FilterInputStream.java:119)
	at java.base/sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read(HttpURLConnection.java:3308)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:279)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:322)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:186)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:183)
	at java.base/java.io.BufferedReader.fill(BufferedReader.java:166)
	at java.base/java.io.BufferedReader.readLine(BufferedReader.java:333)
	at java.base/java.io.BufferedReader.readLine(BufferedReader.java:400)
	at org.glassfish.grizzly.http.server@4.1.0-SNAPSHOT/org.glassfish.grizzly.http.server.NetworkListenerTest.testGracefulShutdownGracePeriodExpired(NetworkListenerTest.java:244)
```
Since JDK 24 GA, if EOF is encountered during BufferedReader#readLine execution, an IOException may be thrown instead of null.

- https://github.com/openjdk/jdk/blob/jdk-24-ga/src/java.base/share/classes/sun/net/www/MeteredStream.java
- https://github.com/openjdk/jdk/blob/6705a9255d28f351950e7fbca9d05e73942a4e27/src/java.base/share/classes/sun/net/www/MeteredStream.java#L57-L59
- https://bugs.openjdk.org/browse/JDK-8335135 - https://github.com/openjdk/jdk/commit/b6ffb442acb4a222f017868433eff213d9b84ed8

```
[ERROR]   AlpnSupportTest.sslToConnectionMapClearTest:53 » Timeout
[ERROR] org.glassfish.grizzly.http2.AlpnSupportTest.sslToConnectionMapClearTest -- Time elapsed: 10.08 s <<< ERROR!
java.util.concurrent.TimeoutException
	at org.glassfish.grizzly.impl.SafeFutureImpl$Sync.innerGet(SafeFutureImpl.java:324)
	at org.glassfish.grizzly.impl.SafeFutureImpl.get(SafeFutureImpl.java:237)
	at org.glassfish.grizzly.http2.AlpnSupportTest.sslToConnectionMapClearTest(AlpnSupportTest.java:53)
```
Internally, javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate) is occurring.

Consideration should be given to avoiding this error in JDK 25 as well.

carryel:
- #2259 
  + Considered the java.io.IOException: Premature EOF error.
  + Changed to Cipher Suites that support TLS 1.3.
- #2258 
  + (Trivial) Added missing license headers.

All builds and testcases passed successfully on JDK 25.

### DIFF
--- a/modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
+++ b/modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -23,6 +24,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import java.io.BufferedReader;
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
@@ -241,7 +243,15 @@ public class NetworkListenerTest {
             Future<HttpServer> gracefulFuture = server.shutdown(1, TimeUnit.SECONDS);
 
             final BufferedReader reader = new BufferedReader(new InputStreamReader(c.getInputStream(), Charsets.UTF8_CHARSET));
-            final String content = reader.readLine();
+            String content;
+            try {
+                // this should result in null in older JDKs(~ JDK 23) or IOException on newer JDKs(JDK 24 GA ~) as the grace period expires
+                content = reader.readLine();
+            } catch (IOException e) {
+                // starting with JDK-24-GA, when EOF is encountered during BufferedReader#readLine() execution,
+                // "java.io.IOException: Premature EOF" is thrown instead of returning null
+                content = null;
+            }
             assertNull(content);
 
             assertNotNull(gracefulFuture.get(5, TimeUnit.SECONDS));
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/util/Parameters.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/util/Parameters.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
--- a/modules/http/src/test/java/org/glassfish/grizzly/http/ParametersTest.java
+++ b/modules/http/src/test/java/org/glassfish/grizzly/http/ParametersTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2011, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
--- a/modules/grizzly/src/main/resources/org/glassfish/grizzly/localization/log.properties
+++ b/modules/grizzly/src/main/resources/org/glassfish/grizzly/localization/log.properties
@@ -1,4 +1,5 @@
 #
+# Copyright (c) 2025 Contributors to the Eclipse Foundation.
 # Copyright (c) 2010, 2017 Oracle and/or its affiliates. All rights reserved.
 #
 # This program and the accompanying materials are made available under the
--- a/modules/http2/src/test/java/org/glassfish/grizzly/http2/AbstractHttp2Test.java
+++ b/modules/http2/src/test/java/org/glassfish/grizzly/http2/AbstractHttp2Test.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -154,11 +155,11 @@ public abstract class AbstractHttp2Test {
                     SSLContextConfigurator sslContextConfigurator = createSSLContextConfigurator();
                     serverSSLEngineConfigurator = new SSLEngineConfigurator(sslContextConfigurator.createSSLContext(true), false, false, false);
 
-                    serverSSLEngineConfigurator.setEnabledCipherSuites(new String[] { "TLS_RSA_WITH_AES_256_CBC_SHA" });
+                    serverSSLEngineConfigurator.setEnabledCipherSuites(new String[] { "TLS_AES_128_GCM_SHA256" });
 
                     clientSSLEngineConfigurator = new SSLEngineConfigurator(sslContextConfigurator.createSSLContext(true), true, false, false);
 
-                    clientSSLEngineConfigurator.setEnabledCipherSuites(new String[] { "TLS_RSA_WITH_AES_256_CBC_SHA" });
+                    clientSSLEngineConfigurator.setEnabledCipherSuites(new String[] { "TLS_AES_128_GCM_SHA256" });
                 }
             }
         }


pzygielo (approved):
If this PR fixes some issue - please exercise [Using keywords in issues and pull requests](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) or link it in _Development_ section to automate closing such item.

Also see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests:
> If you reference an issue, pull request, or discussion in a list, the reference will unfurl to show the title and state instead.

Thus difference in formatting
#2260 (not in list)
and
- #2260 


(So funny. Unfurling seems to work in preview of this very comment but not in its final view. Very funny.)

image
image

pzygielo:
#2260

- #2260

carryel:
@pzygielo As always, thank you so much for the information. I've been curious about this. I'll read it and apply it from now on.

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
```diff
@@ -241,7 +243,14 @@ public void testGracefulShutdownGracePeriodExpired() throws Exception {
             Future<HttpServer> gracefulFuture = server.shutdown(1, TimeUnit.SECONDS);
 
             final BufferedReader reader = new BufferedReader(new InputStreamReader(c.getInputStream(), Charsets.UTF8_CHARSET));
-            final String content = reader.readLine();
+            String content;
+            try {
+                // null is expected here as the grace period will expire
+                content = reader.readLine();
+            } catch (IOException e) {
+                // or io exception("Premature EOF") is expected here in some high JDKs
```
OndroMih: Do you know why? This comment is very vague, ideally it should explain why and since which Java version.

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
```diff
@@ -241,7 +243,14 @@ public void testGracefulShutdownGracePeriodExpired() throws Exception {
             Future<HttpServer> gracefulFuture = server.shutdown(1, TimeUnit.SECONDS);
 
             final BufferedReader reader = new BufferedReader(new InputStreamReader(c.getInputStream(), Charsets.UTF8_CHARSET));
-            final String content = reader.readLine();
+            String content;
+            try {
+                // null is expected here as the grace period will expire
```
OndroMih: This comment doesn't make much sense. Is `null` expected on this line? Wouldn't it result in a nullptr exception?
Or it's expected that reader.readLine() throws IOException and then it's OK to set the content to `null`?

I think what you want to say is that `content == null` means the grace period expired. And on newer JDKs the behavior changes - an exception is thrown if grace period expires, instead of returning null.

What about this comment:

```suggestion
                // this should result in null in older JDKs or IOException on newer JDKs as the grace period expires 
```

modules/http/src/test/java/org/glassfish/grizzly/http/ParametersTest.java
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
```
OndroMih, line 2: Reason: 
File was modified earlier without updating the copyright: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2258/files

modules/http/src/main/java/org/glassfish/grizzly/http/util/Parameters.java
```diff
@@ -1,4 +1,5 @@
 /*
```
OndroMih, line 1: Reason: 
File was modified earlier without updating the copyright: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2258/files

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
pzygielo: - Isn't this described in linked https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2259?

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
carryel: Perhaps I should add a little more explanation. I understand why you were confused. I checked the actual JDK source and noticed that the behavior changed (JDK17 vs. JDK25). However, I didn't find any notices or changelogs regarding this behavior change. However, I was only aware that security and network protocols have become more stringent since JDK 25.

I will try to provide more detailed explanations and supplement the content later.

pzygielo: > Perhaps I should add a little more explanation. I understand why you were confused. I checked the actual JDK source and noticed that the behavior changed (JDK17 vs. JDK25). 

Maybe this https://bugs.openjdk.org/browse/JDK-8335135 - https://github.com/openjdk/jdk/commit/b6ffb442acb4a222f017868433eff213d9b84ed8

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
OndroMih: Thanks. I think we should clarify in the comment, something like "since Java 24 because of stricter enforcement of networking protocols"

modules/http-server/src/test/java/org/glassfish/grizzly/http/server/NetworkListenerTest.java
carryel: @OndroMih The `testGracefulShutdownGracePeriodExpired` test does not completely test whether the listener is gracefully shut down.
Graceful shutdown means shutting down when there are no currently active or in-progress http requests and all processing has been completed.

This test generates a shutdown event, delivers it to each listener and registered filter, maintains an unfinished state, and assumes that forced shutdown occurs successfully.

Looking at the test case, Grizzly generates a shutdown event and has a grace period of 1 second.
Grizzly handler first sends a response containing http status and header, waits 2 seconds (expects grace period to expire), and then writes msg.
The core of the test appears to be testing whether or not the shutdown proceeds after the grace period expires after the shutdown event is delivered/propagated through whether or not this msg is received.
In other words, we are verifying that after expiration, shutdown occurs and no msg is received. It seems that the client side probably assumed an EOF situation.

BufferedReader#readLine() is explained as follows:
```
Returns:
A String containing the contents of the line, not including any line-termination characters, or null if the end of the stream has been reached without reading any characters
```
Additionally, there was no change to the intent of returning null in EOF situations.
https://github.com/openjdk/jdk/blob/2c07214d7c075da5dd4a4e872aef29f58cef2bae/src/java.base/share/classes/java/io/BufferedReader.java#L332-L338

However, like the part linked by @pzygielo above, the fill() behavior has been changed to throw an IOException at EOF before returning null.

In a situation where null is returned in the old JDK and an IOException occurs in the new JDK, I briefly thought about changing the test case in a different way to prevent such changes in the JDK, but for comparison purposes, I tried to maintain the existing source as much as possible, so it ended up like this.
I thought it would be less confusing to leave at least a hint rather than just applying the modifications, and I thought it was a test case without much meaning, so I made comments too roughly, which may have caused more confusion.

As you advised, I will submit an edit to leave a more clear comment. :)

OndroMih: Thanks @carryel, makes perfect sense now! I was maybe a little too pedantic. i just thought it’s good to leave a clear comment in the code for people that come across it in the future :)

OndroMih (approved):

PR #2270, opened 2026-01-04
Fixed plugin management of the helper plugin and few warnings

arjantijms:
#2189 removed the groupId by mistake.
Maven also warned about implicit versions of few maven plugins.


pom.xml
```diff
@@ -209,6 +199,11 @@
                         </instructions>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>build-helper-maven-plugin</artifactId>
```
pzygielo: Isn't this already managed in parent?

OndroMih (approved):

dmatej:
Wouldn't be better to just rebase the original PR? Git can handle that, we both would be authors of the commit and it would not need the additional merge commit. ;)
I have rebased the original PR now.
This PR now improves the buildhelper plugin settings, if you agree, I can rebase it to my again.

dmatej (changes requested):
```
Warning:  
Warning:  Some problems were encountered while building the effective model for org.glassfish.grizzly:grizzly-framework:bundle:5.0.0-SNAPSHOT
Warning:  'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing. @ line 90, column 21
Warning:  
Warning:  Some problems were encountered while building the effective model for org.glassfish.grizzly:grizzly-framework-monitoring:bundle:5.0.0-SNAPSHOT
Warning:  'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing. @ line 103, column 21
Warning:  
Warning:  Some problems were encountered while building the effective model for org.glassfish.grizzly:grizzly-http-monitoring:bundle:5.0.0-SNAPSHOT
Warning:  'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing. @ line 100, column 21
Warning:  
Warning:  Some problems were encountered while building the effective model for org.glassfish.grizzly:grizzly-http-server-monitoring:bundle:5.0.0-SNAPSHOT
Warning:  'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing. @ line 101, column 21
Warning:  
Warning:  It is highly recommended to fix these problems because they threaten the stability of your build.
Warning:  
Warning:  For this reason, future Maven versions might no longer support building such malformed projects.
```

dmatej:
The BOM is a trap here, it is a module of the parent and modules inherit from the parent, not from BOM. Probably it would be better to remove all plugins from BOM and move them to parent, while BOM would keep just the dependency management.

pzygielo:
> The BOM is a trap here, it is a module of the parent and modules inherit from the parent,

Which declares bom as parent.

> not from BOM. 

So isn't boms/bom as the top parent (just before superpom)?

dmatej:
Aha, I missed this one. Then why Maven complained? Always in bundle ... I will recheck ...

pzygielo:
> I will recheck ...

Doing the same :smile:

dmatej:
I think I have seen it somewhere already, having bom as a module while depending on it as a parent is quite dirty idea ... perhaps this is the problem. Is this still supported with Maven4?

pzygielo:
> Aha, I missed this one. Then why Maven complained? Always in bundle ... I will recheck ...


https://github.com/eclipse-ee4j/glassfish-grizzly/blob/9dfb3edfac2394af564e483dc758e51b24b0ee89/boms/bom/pom.xml#L196-L199

is missing explicit `groupId`.

dmatej:
Haha, great! :-)
I am testing the fixed build now ...

pzygielo:
- https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2189

dmatej (approved):

pzygielo (approved):

PR #2276, opened 2026-01-15
Fixed MultiEndPointPoolTest stability

dmatej:
- Closing connections is asynchronous
- The testLocalAddress test was leaking connection but the class did not check

### DIFF
--- a/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
+++ b/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2025, 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2013, 2017 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -18,6 +18,7 @@
 package org.glassfish.grizzly.connectionpool;
 
 import java.io.IOException;
+import java.lang.System.Logger;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.SocketAddress;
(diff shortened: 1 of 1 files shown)


OndroMih (approved):

carryel:
@dmatej @OndroMih Grizzly nightly build has been failing for a while now.
https://ci.eclipse.org/glassfish/view/Grizzly/job/Grizzly-master_nightly/2903/testReport/

Out of curiosity, I looked into it a bit.

1. MultiEndpointPool and SingleEndpointPool return immediately when closing a connection without receiving the close result. This is because they use the void Connection#closeSilently method.

2. When the pool closes, it only closes ready connections (those that have been released and managed). It does not close connections that are in use. In other words, in the testSingleEndpointClose test, the key2 connection is not explicitly closed in the test case.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/df6c52bc15588deae0716c0804f14b77fa3e201d/extras/connection-pool/src/main/java/org/glassfish/grizzly/connectionpool/SingleEndpointPool.java#L861-L865

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/df6c52bc15588deae0716c0804f14b77fa3e201d/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java#L285

Therefore, due to the above two characteristics, even if the pool closes, there's no guarantee that the connection is immediately closed. Depending on the environment, the closing process may require some time.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/df6c52bc15588deae0716c0804f14b77fa3e201d/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java#L105

I'll think a little more about how to resolve this situation and submit a separate PR if necessary.

OndroMih:
@carryel , do I understand correctly that it's a problem in tests, not in production code?

> I'll think a little more about how to resolve this situation and submit a separate PR if necessary.

If you can find a way to fix it and submit a PR, that would be brilliant :)

carryel:
@OndroMih That's right. This is the intended behavior of the pool's close operation.
Of course, these connections will be cleaned up during the selector shutdown process via transport.shutdownNow(), but this is also not a blocking close.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/df6c52bc15588deae0716c0804f14b77fa3e201d/modules/grizzly/src/main/java/org/glassfish/grizzly/nio/SelectorRunner.java#L204-L207

Yes, I will try to patch the test case as simply as possible.

PR #2277, opened 2026-01-20
Fixed MultiEndPointPoolTest stability(2)

carryel:
- https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2276#issuecomment-3767473685
- tearDown() ensures that all open connections are closed.
- Checking whether connections are properly closed after the pool is closed is verified within a separate testSingleEndpointClose() test.

### DIFF
--- a/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
+++ b/extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
@@ -23,8 +23,11 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.SocketAddress;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
@@ -44,7 +47,9 @@ import org.junit.Before;
 import org.junit.Test;
 
 import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.INFO;
 import static java.util.Collections.newSetFromMap;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
```diff
@@ -102,7 +104,20 @@ public void tearDown() throws IOException {
...
-            assertFalse("Pool is closed, but connection is still open: " + connection, connection.isOpen());
+            if (connection.isOpen()) {
+                // For connections not managed by the Pool (not returned to the Pool), the connection may persist even if the Pool closes.
+                // These connections are terminated in a separate thread during the Selector shutdown process when the Transport is shut down.
+                // Therefore, connection's termination cannot be assumed to be immediate; it may take some time to complete depending on the timing.
+                LOG.log(INFO,
+                        "Pool is closed, but connection is still open. We will explicitly close that connection: {0}",
+                        connection);
+                final GrizzlyFuture<Closeable> closeFuture = connection.close();
+                try {
```
OndroMih: Would be better if all connections are closed without waiting, in parallel, futures are stored in a list and then waiting happens in another loop later.

extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
```diff
@@ -102,7 +104,20 @@ public void tearDown() throws IOException {
...
+                    closeFuture.get(1, SECONDS);
```
OndroMih: Would be better if all connections are closed without waiting, in parallel, futures are stored in a list and then waiting happens in another loop later.

carryel: @OndroMih Thanks for the feedback. I've applied your advice with a few tricks. :)

extras/connection-pool/src/test/java/org/glassfish/grizzly/connectionpool/MultiEndPointPoolTest.java
```diff
@@ -101,8 +105,28 @@ public void tearDown() throws IOException {
...
+            if (connection.isOpen()) {
+                // For connections not managed by the Pool (not returned to the Pool), the connection may persist even if the Pool closes.
+                // These connections are terminated in a separate thread during the Selector shutdown process when the Transport is shut down.
+                // Therefore, connection's termination cannot be assumed to be immediate; it may take some time to complete depending on the timing.
+                LOG.log(INFO,
+                        "Pool is closed, but connection is still open. We will explicitly close that connection: {0}",
+                        connection);
+                closeFutures.add(CompletableFuture.runAsync(() -> {
+                    try {
+                        connection.close().get();
```
OndroMih: Better to add timeout here, when calling get method, just in case

OndroMih (approved):
Very well! Approved, with one more suggestion

Issue #2278, opened 2026-01-20
5.0.0-SNAPSHOT referenced by grizzly-project-5.0.0.pom


pzygielo:
https://repo.maven.apache.org/maven2/org/glassfish/grizzly/grizzly-project/5.0.0/grizzly-project-5.0.0.pom:

```xml
    <parent>
        <groupId>org.glassfish.grizzly</groupId>
        <artifactId>grizzly-bom</artifactId>
        <version>5.0.0-SNAPSHOT</version>
        <relativePath>boms/bom/pom.xml</relativePath>
    </parent>
```
👀 

-----

- https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2275

OndroMih:
@dmatej

pzygielo:
- This makes some tools to fail for missing metadata, like openrewrite in https://github.com/eclipse-ee4j/openmq/pull/2809.

pzygielo:
And with access to `central-portal-snapshots` blocked regular maven build does not like it as well:
```
Could not collect dependencies for project ...
[ERROR] Failed to read artifact descriptor for org.glassfish.grizzly:grizzly-portunif:jar:5.0.0
[ERROR] 	Caused by: The following artifacts could not be resolved: org.glassfish.grizzly:grizzly-bom:pom:5.0.0-SNAPSHOT (absent): Could not find artifact org.glassfish.grizzly:grizzly-bom:pom:5.0.0-SNAPSHOT in central-portal-snapshots (... blocked...)
```

pzygielo:
This is perhaps the result of the same release job, that top-level `pom.xml` in https://github.com/eclipse-ee4j/glassfish-grizzly/tree/5.0.0 has version updated, but not version on parent:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/bce1a0dd1f0e3cf236369bf03dfe2184092d19cd/pom.xml#L24-L34

while the bom-bom parent is not even touched:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/bce1a0dd1f0e3cf236369bf03dfe2184092d19cd/boms/bom/pom.xml#L33


😞

pzygielo:
@arjantijms Is there a plan to re-release something like 5.0.1 soon to address this issue and make Grizzly 5 fully usable?

carryel:
@arjantijms I'm also curious if the release was made using the Jenkins job below.

[...]

PR #2282, opened 2026-01-27
Fixed initialization of Async Servlet ReadListener/WriteListener callbacks

### LINKED ISSUE #2281: Async Servlet ReadListener/WriteListener callbacks never fire unless isReady() is called
### Summary
When using non-blocking servlet IO, `ReadListener.onDataAvailable()` and `WriteListener.onWritePossible()` are never invoked unless the application manually calls `ServletInputStream.isReady()`/`ServletOutputStream.isReady()`. This causes async servlets to hang and never respond.

### Steps to Reproduce
1. Create an async servlet and call `request.startAsync()`.
2. Call `request.getInputStream().setReadListener(...)` or `response.getOutputStream().setWriteListener(...)`.
3. Do NOT call `isReady()` in the servlet.
4. Send a request with a body (for read) or attempt to write a response in `onWritePossible()` (for write).

### Expected
`ReadListener.onDataAvailable()` and `WriteListener.onWritePossible()` are invoked by the container once IO is possible, per Servlet async IO behavior.

### Actual
Callbacks are never invoked unless `isReady()` is explicitly called by the app. Responses hang when using `WriteListener`.

### Notes
The servlet wrapper registers Grizzly handlers only when `isReady()` is called. In practice many apps rely on `setReadListener`/`setWriteListener` to start notifications without calling `isReady()`.

### Impact
gRPC servlet adapters (and other async servlet frameworks) hang because they expect the container to notify listeners once IO is possible.

### Fix
Register the Grizzly handler immediately in `setReadListener`/`setWriteListener`.

```diff
diff --git a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java
index 4372393..917b104 100644
--- a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java
+++ b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java
@@ -188,6 +188,7 @@ public class ServletInputStreamImpl extends ServletInputStream {
 
         readHandler = new ReadHandlerImpl(readListener);
         hasSetReadListener = true;
+        inputStream.notifyAvailable(readHandler);
     }
 
     class ReadHandlerImpl implements ReadHandler {

```

```diff
diff --git a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java
index 73cae22..095bb37 100644
--- a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java
+++ b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java
@@ -141,6 +141,7 @@ public class ServletOutputStreamImpl extends ServletOutputStream {
 
         writeHandler = new WriteHandlerImpl(writeListener);
         hasSetWriteListener = true;
+        outputStream.notifyCanWrite(writeHandler);
     }
 
     void recycle() {
```

```java
@WebServlet(urlPatterns = "/async", asyncSupported = true)
public class AsyncUploadServlet extends HttpServlet {

    @Override
    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
            throws IOException {

        AsyncContext asyncContext = req.startAsync();
        asyncContext.setTimeout(0); // optional: disable timeout

        ServletInputStream input = req.getInputStream();
        ByteArrayOutputStream buffer = new ByteArrayOutputStream();

        input.setReadListener(new ReadListener() {

            @Override
            public void onDataAvailable() throws IOException {
                byte[] bytes = new byte[1024];
                int len;

                while (input.isReady() && (len = input.read(bytes)) != -1) {
                    buffer.write(bytes, 0, len);
                }
            }

            @Override
            public void onAllDataRead() throws IOException {
                HttpServletResponse response =
                        (HttpServletResponse) asyncContext.getResponse();

                response.setContentType("text/plain");
                ServletOutputStream output = response.getOutputStream();

                byte[] responseBytes =
                        ("Received " + buffer.size() + " bytes").getBytes();

                output.setWriteListener(new WriteListener() {

                    int offset = 0;

                    @Override
                    public void onWritePossible() throws IOException {
                        while (output.isReady() && offset < responseBytes.length) {
                            output.write(responseBytes[offset++]);
                        }

                        if (offset >= responseBytes.length) {
                            asyncContext.complete();
                        }
                    }

                    @Override
                    public void onError(Throwable t) {
                        t.printStackTrace();
                        asyncContext.complete();
                    }
                });
            }

            @Override
            public void onError(Throwable t) {
                t.printStackTrace();
                asyncContext.complete();
            }
        });
    }
}
```

carryel:
- https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2281
- The behavior of Read/Write listener callbacks requiring explicit isReady() calls to ServletInputStream/ServletOutputStream to be registered has been modified to allow initial registration when registering the listener.
- According to the Servlet spec, isReady() in blocking mode should always be true, but the behavior of throwing an IllegalStateException has been modified.
- Related test cases have been added.

### DIFF
--- a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java
+++ b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletInputStreamImpl.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
@@ -146,10 +147,6 @@ public class ServletInputStreamImpl extends ServletInputStream {
      */
     @Override
     public boolean isReady() {
-        if (!hasSetReadListener) {
-            throw new IllegalStateException(LogMessages.WARNING_GRIZZLY_HTTP_SERVLET_INPUTSTREAM_ISREADY_ERROR());
-        }
-
         if (!prevIsReady) {
             return false;
         }
@@ -187,6 +184,18 @@ public class ServletInputStreamImpl extends ServletInputStream {
         }
 
         readHandler = new ReadHandlerImpl(readListener);
+
+        // Initial installation of ReadListener that will be notified
+        if (!inputStream.isReady()) {
+            prevIsReady = false;
+        }
+        IS_READY_SCOPE.set(Boolean.TRUE);
+        try {
+            inputStream.notifyAvailable(readHandler);
+        } finally {
+            IS_READY_SCOPE.remove();
+        }
+
         hasSetReadListener = true;
     }
 
--- a/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java
+++ b/modules/http-servlet/src/main/java/org/glassfish/grizzly/servlet/ServletOutputStreamImpl.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
@@ -98,10 +99,6 @@ public class ServletOutputStreamImpl extends ServletOutputStream {
      */
     @Override
     public boolean isReady() {
-        if (!hasSetWriteListener) {
-            throw new IllegalStateException(LogMessages.WARNING_GRIZZLY_HTTP_SERVLET_OUTPUTSTREAM_ISREADY_ERROR());
-        }
-
         if (!prevIsReady) {
             return false;
         }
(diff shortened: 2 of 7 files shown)


carryel:
@pzygielo @avpinchuk @OndroMih @arjantijms @dmatej Could someon please review this issue?

Grizzly has always registered a callback listener only when isReady() is called (if input/output processing is not possible when isReady() is called). Existing test cases and samples also assume this behavior. This behavior likely stems from the fact that processing input/output with a separate callback is generally inefficient when the current thread and input/output are ready.

However, this behavior differs from that of other products. Even if the current thread and intput/output are ready, if a callback listener is registered, processing will typically be done with that callback.

Additionally, the contributor who requested this improvement hopes that this feature will be included in 5.0.1. Will this also be considered for the 5.0.1 release?

- It would be great if this fix is ​​included when https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2278 is resolved.

avpinchuk (approved):
LGTM. Sorry for the late response.

PR #2283, opened 2026-02-02
Update README with branch information for Grizzly

OndroMih:

PR #2284, opened 2026-02-04
Improved README.md

carryel:
- Updated the Prerequisites.
- Removed the non-existent https://eclipse-ee4j.github.io/grizzly/dependencies.html and summarized some of the content from what is believed to be the original link, https://javaee.github.io/grizzly/dependencies.html.
  - All JAR and bundle information has been organized into a table. It is quite long, so it has been divided into a separate section.
  - Added instructions for importing using the BOM.
- https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2283#issuecomment-3839930531

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,7 @@ help developers to build scalable and robust servers using NIO as well
 as offering extended framework components: Web Framework (HTTP/S),
 WebSocket, Comet, and more!
 
-
 ### Versions and Branches
-
 
 - `main` : This is the main development branch, currently for 5.0.x (Jakarta EE 11, Java 21+). (latest release is 5.0.0)
 - `4.1.x` : This is the sustaining branch for 4.1.x (Jakarta EE 11 with Java 17 support). (no release yet)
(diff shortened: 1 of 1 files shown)


avpinchuk:
@carryel, what about this link https://eclipse-ee4j.github.io/glassfish-grizzly/dependencies.html?

OndroMih (approved):
Very nice and useful!

OndroMih:
Hi, @carryel , this info is very useful, I also learned a lot more about Grizzly than I knew before.

Maybe later we could copy the info also to the website in the gh-pages branch: https://github.com/eclipse-ee4j/glassfish-grizzly/tree/gh-pages

carryel:
@avpinchuk @OndroMih Oh...
I missed the updated link https://eclipse-ee4j.github.io/glassfish-grizzly/dependencies.html. Had I known about it, I would have simply updated the link (which was incorrect). (The content is the same in both the old and new links, and in the README.)

However, now that I've written it, I think it might be helpful to briefly explain the usage through the README without having to navigate externally. It also makes it easier to update future updates. So, I think it might be a good idea to add the link (https://eclipse-ee4j.github.io/glassfish-grizzly/dependencies.html) as a reference.

It can also be referenced at https://github.com/eclipse-ee4j/glassfish-grizzly/tree/gh-pages in the future.

As always, thank you for your feedback.

avpinchuk (approved):

OndroMih:
Now that you shared the correct link, I see that the website already has the info. And the dependencies page is linked from the main page, in the Download section. So there's no need to update the pages in gh-pages, unless you added some more info into the README.

It's OK to duplicate the info in the README, while it's also useful to have a link to the dependencies page on the web for reference.

PR #2288, opened 2026-04-08
Optimize object holder

avpinchuk:
In this patch:
* Remove unused code
* Decrease volatile reads
* Use weaker memory access operations

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
@@ -17,6 +17,10 @@
 
 package org.glassfish.grizzly.utils;
 
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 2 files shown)


modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
```diff
@@ -73,57 +59,41 @@ public String toString() {
...
     public static abstract class LazyHolder<E> extends Holder<E> {
-        private volatile boolean isSet;
-        private E value;
 
-        @Override
-        public final E get() {
-            if (isSet) {
-                return value;
-            }
+        private static final Object UNINITIALIZED = new Object();
```
carryel, line 63: Look good to me.

I have a few questions I would like to confirm.

1. Is there a reason you explicitly specified `UNINITIALIZED` object instead of comparing with null?

modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
```diff
@@ -73,57 +59,41 @@ public String toString() {
...
-            return getInt();
-        }
-
-        public abstract int getInt();
-    }
-
-    public static abstract class LazyIntHolder extends IntHolder {
-        private volatile boolean isSet;
-        private int value;
+        private volatile Object value = UNINITIALIZED;
```
carryel, line 76: 2. I understand that this part does not strictly need to be declared as volatile, and I believe there are no performance issues even with volatile. I am curious if you declared it with this in mind.

modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
```diff
@@ -73,57 +59,41 @@ public String toString() {
...
+                lock.lock();
+                try {
+                    localValue = VALUE.getAcquire(this);
+                    if (localValue == UNINITIALIZED) {
+                        localValue = evaluate();
+                        VALUE.setRelease(this, localValue);
+                    }
+                } finally {
+                    lock.unlock();
                 }
```
carryel, line 92: 3. Although there are pros and cons, have you considered a lock-free version as follows? Using the lock-free version does result in unnecessary initialization that would otherwise be wasted.

```
        public final E get() {
            Object localValue = VALUE.getAcquire(this);
            if (localValue == UNINITIALIZED) {
                localValue = evaluate();
                Object witness = VALUE.compareAndExchange(this, UNINITIALIZED, localValue);
                if (witness != null) {
                    localValue = witness;
                }
            }
            return (E) localValue;
        }
```

carryel (approved):

modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
avpinchuk: Using `UNINITIALIZAD` provides null safety if `null` is a valid value. For example, calling socket.getLocalAddress() might return `null`. Otherwise, we will try to initialize the holder every time in this case. This is a replacement for the `isSet` field.

modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
avpinchuk: This must be declared as a `volatile` to ensure the `UNINITIALIZED` value is visible.

avpinchuk: There is no performance penalty here, since other than assigning the initial value, all access is done via VarHandle with acquire/release semantics.

This provides a performance boost on platforms with a weakened memory model, such as ARM.

modules/grizzly/src/main/java/org/glassfish/grizzly/utils/Holder.java
avpinchuk: I thought about that. But where the holder is used, socket operations are used, which ultimately syscall the kernel. It's cheaper to use a lock than to make several (potentially many) expensive syscalls.

avpinchuk: The `localValue` variable is needed to reduce the number of volatile reads.

PR #2290, opened 2026-04-12
Improve options processing in the `TCPNIOTransport`

avpinchuk:
What's changed:

* Do not set `SO_REUSEADDR` and `SO_REUSEPORT` options if the socket is bound because behavior is undefined.
* Do not set `SO_LINGER` and `SO_TIMEOUT` options for a non-blocking socket because behavior is undefined.
* If setting an option reveals that the socket is in an invalid state or closed, do not set the remaining options to avoid expensive syscalls. This may happen, for example, when the RST segment received.
* Setting socket options using the `setOption()` method (NIO.2 way).

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
@@ -23,6 +23,7 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketAddress;
+import java.net.SocketException;
 import java.net.StandardSocketOptions;
 import java.nio.channels.ClosedByInterruptException;
 import java.nio.channels.SelectableChannel;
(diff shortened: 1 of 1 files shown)


modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
```diff
@@ -700,96 +701,117 @@ public ProcessorSelector getProcessorSelector() {
...
+                final Socket socket = socketChannel.socket();
+
+                if (socket.isBound()) {
+                    return;
+                }
 
                 final boolean reuseAddress = tcpNioTransport.isReuseAddress();
                 try {
-                    socket.setReuseAddress(reuseAddress);
+                    socket.setOption(StandardSocketOptions.SO_REUSEADDR, reuseAddress);
```
carryel, line 726: I am curious why setOption() is used even though a direct, explicit interface is provided. It does seem that there is no difference in the JDK implementations, though.

modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
```diff
@@ -700,96 +701,117 @@ public ProcessorSelector getProcessorSelector() {
...
-                    socket.setKeepAlive(keepAlive);
+                    socket.setOption(StandardSocketOptions.TCP_NODELAY, tcpNoDelay);
                 } catch (IOException e) {
-                    LOGGER.log(Level.FINE, LogMessages.FINE_GRIZZLY_SOCKET_KEEPALIVE_EXCEPTION(keepAlive, socket), e);
+                    LOGGER.log(Level.FINE, e, () -> LogMessages.FINE_GRIZZLY_SOCKET_TCPNODELAY_EXCEPTION(tcpNoDelay, socket));
+                    return;
                 }
 
-                final boolean tcpNoDelay = tcpNioTransport.isTcpNoDelay();
+                if (!socketChannel.isBlocking()) {
```
carryel, line 794: What likely affects the operation is the SO_LINGER option setting on the next line. Generally, SO_LINGER is meaningless in non-blocking environments, but wouldn't `setSoLinger(true, 0)`, which sets a zero timeout, be meaningful even in non-blocking?

modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
avpinchuk: @carryel, for consistency because we still use SO_REUSEPORT.

modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java
avpinchuk: I tried to follow the specification in the javadoc for the`SO_LINGER` option:

> This socket option is intended for use with sockets that are configured in [blocking](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/nio/channels/SelectableChannel.html#isBlocking()) mode only. **The behavior of the close method when this option is enabled on a non-blocking socket is not defined**.

carryel (approved):

dmatej (approved):

carryel:
@avpinchuk This PR's status is not ready, right? If there are no further modifications, please change the status.

avpinchuk:
@carryel I still need to clarify some minor points about error handling.

dmatej (approved):

PR #2291, opened 2026-04-15
Fix #2286 HTTP/2 communication may return HTTP 400

### LINKED ISSUE #2286: HTTP/2 communication may return HTTP 400
## Problem Description
When communicating over HTTP/2, an HTTP 400 status in HTTP/1.1 format may be returned even though the communication is performed using HTTP/2.
This issue is caused by the validation logic of the HTTP/2 client connection preface (PRI and SETTINGS frames) in Grizzly.
Normally, the PRI and SETTINGS frames are processed together within a single READ event. However, depending on network conditions, they may be processed separately across multiple READ events.
In such cases, validation logic intended only for the PRI frame is also applied to the SETTINGS frame, causing the request to be treated as invalid and resulting in an HTTP 400 response.

## Environment
* Server:GlassFish 8.0.0(Grzzly 5.0.0)
* JDK: openjdk version "21.0.8"
* OS : Windows 11

## Steps to reproduce by glassfish
1. Run GlassFish in a configuration without a front-end web server.
2. Send an HTTP/2 request directly to GlassFish.
3. Ensure that the HTTP/2 client connection preface is received in separate READ events.
4. Under this condition, GlassFish returns HTTP status 400.
The following script reproduces the issue from a Windows client environment.
Please change the file extension from .txt to .ps1 before using the script.
[repro-h2c-400.txt](https://github.com/user-attachments/files/26736318/repro-h2c-400.txt)

```
> powershell -ExecutionPolicy Bypass -File .\repro-h2c-400.ps1
[Attempt 1] HTTP_400 = DETECTED

--- 400 RESPONSE ---
HTTP/1.1 400 Bad Request
Server: Eclipse GlassFish 8.0.0
X-Powered-By: Servlet/6.1 JSP/4.0(Eclipse GlassFish 8.0.0 Java/Eclipse Adoptium/21)
Date: Wed, 15 Apr 2026 04:38:58 GMT
Connection: close
Transfer-Encoding: chunked


--- END RESPONSE ---


HTTP_400_SUMMARY: DETECTED=1/1, NOT_DETECTED=0/1
FINAL_RESULT: HTTP_400_DETECTED
```

## Impact of Issue
- HTTP/2 requests may be rejected with HTTP status 400.

11rx4f:
* Fixes #2286 

Fixes the HTTP/2 connection preface validation to prevent PRI-specific validation from being executed multiple times.

Signed-off-by: 11rx4f [ryosuke.okada@fujitsu.com](mailto:ryosuke.okada@fujitsu.com)

### DIFF
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
@@ -291,8 +291,7 @@ public class Http2ServerFilter extends Http2BaseFilter {
             }
         }
 
-        final Buffer framePayload;
-        if (!http2Session.isHttp2InputEnabled() && !http2State.isPriReceived()) { // Preface is not received yet
+        if (!http2Session.isHttp2InputEnabled()) { // Preface is not received yet
 
             if (http2State.isHttpUpgradePhase()) {
                 // It's plain HTTP/1.1 data coming with upgrade request
@@ -303,7 +302,10 @@ public class Http2ServerFilter extends Http2BaseFilter {
 
                 return ctx.getInvokeAction();
             }
+        }
 
+        final Buffer framePayload;
+        if (!http2State.isPriReceived()) { // PRI is not received yet
             final HttpRequestPacket httpRequest = (HttpRequestPacket) httpHeader;
 
             // PRI message hasn't been checked
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
@@ -292,7 +292,7 @@ public class Http2ServerFilter extends Http2BaseFilter {
         }
 
         final Buffer framePayload;
-        if (!http2Session.isHttp2InputEnabled()) { // Preface is not received yet
+        if (!http2Session.isHttp2InputEnabled() && !http2State.isPriReceived()) { // Preface is not received yet
 
             if (http2State.isHttpUpgradePhase()) {
                 // It's plain HTTP/1.1 data coming with upgrade request
@@ -312,6 +312,7 @@ public class Http2ServerFilter extends Http2BaseFilter {
                     // Not enough PRI content read
                     return ctx.getStopAction(httpContent);
                 }
+                http2State.setPriReceived(true);
             } catch (Exception e) {
                 httpRequest.getProcessingState().setError(true);
                 httpRequest.getProcessingState().setKeepAlive(false);
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
@@ -69,7 +69,7 @@ class Http2State {
 
     private Http2Session http2Session;
 
-    private volatile boolean isPriReceived;
+    private boolean isPriReceived;
 
     private boolean isClientHttpUpgradeRequestFinished;
     private boolean isClientPrefaceSent;
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
@@ -69,6 +69,8 @@ class Http2State {
 
     private Http2Session http2Session;
 
+    private volatile boolean isPriReceived;
+
     private boolean isClientHttpUpgradeRequestFinished;
     private boolean isClientPrefaceSent;
 
@@ -158,6 +160,14 @@ class Http2State {
     void setHttp2Session(final Http2Session http2Session) {
         this.http2Session = http2Session;
         this.http2Session.http2State = this;
+    }
+
+    boolean isPriReceived() {
+        return isPriReceived;
+    }
+
+    void setPriReceived(boolean isPriReceived) {
+        this.isPriReceived = isPriReceived;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


carryel:
@11rx4f Thank you for submitting the PR! Could you possibly add relevant test cases for this issue as well?

11rx4f:
@carryel  Sure, I’ll add the relevant test cases.

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
```diff
@@ -69,6 +69,8 @@ static void remove(final Connection connection) {
 
     private Http2Session http2Session;
 
+    private volatile boolean isPriReceived;
```
pzygielo, line 72: No other field here is `volatile`. Why this one should be?

carryel: Additionally, have you considered putting that property into Http2Session? The reason I thought of this is that Http2Session already manages similar states, such as isPrefaceReceived.

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
```diff
@@ -292,7 +292,7 @@ public NextAction handleRead(final FilterChainContext ctx) throws IOException {
         }
 
         final Buffer framePayload;
-        if (!http2Session.isHttp2InputEnabled()) { // Preface is not received yet
+        if (!http2Session.isHttp2InputEnabled() && !http2State.isPriReceived()) { // Preface is not received yet
```
carryel, line 295: If http2State.isPriReceived() is false, wouldn't http2Session.isHttp2InputEnabled() always be false as well?

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
11rx4f: @pzygielo I made this field volatile for better safety and visibility in a potential concurrent scenario.
Do you think it would also make sense to apply volatile to the other fields?

@carryel I did consider putting this property into Http2Session.
However, since this flag is intended to suppress PRI re-validation before the HTTP/2 session is established,
I thought it might fit more naturally in Http2State, which handles connection-phase state.

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
11rx4f: Thanks for pointing that out. When isPriReceived is false, isHttp2InputEnabled is also false in this implementation. They still represent different states: isPriReceived is about PRI validation completion, while isHttp2InputEnabled is about full handshake readiness.

11rx4f (commented):
FYI, this review was started unintentionally while replying.
Please treat the comments as regular discussion points.

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
pzygielo: > Do you think it would also make sense to apply volatile to the other fields?

No. To drop it from this one.

11rx4f:
I’ve added the relevant test cases.

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2ServerFilter.java
carryel: I would like to recommend two things. Please take a look at my suggestion as well.

1. We have separated the logic for Preface received and the logic for PRI checked. Since Preface is always 'not received' if PRI is not checked, we have clearly separated these logics.

```
        if (!http2Session.isHttp2InputEnabled()) { // Preface is not received yet
            if (http2State.isHttpUpgradePhase()) {
                // It's plain HTTP/1.1 data coming with upgrade request
                if (httpContent.isLast()) {
                    http2State.setDirectUpgradePhase(); // expecting preface
                    enableOpReadNow(ctx);
                }
                return ctx.getInvokeAction();
            }
        }
        final Buffer framePayload;
        if (!http2State.isPriReceived()) { // PRI is not checked yet
            final HttpRequestPacket httpRequest = (HttpRequestPacket) httpHeader;

            // PRI message hasn't been checked
            try {
                if (!checkPRI(httpRequest, httpContent)) {
                    // Not enough PRI content read
                    return ctx.getStopAction(httpContent);
                }
                http2State.setPriReceived(true);
            } catch (Exception e) {
                httpRequest.getProcessingState().setError(true);
                httpRequest.getProcessingState().setKeepAlive(false);

                final HttpResponsePacket httpResponse = httpRequest.getResponse();
                httpResponse.setStatus(HttpStatus.BAD_REQUEST_400);
                ctx.write(httpResponse);
                connection.closeSilently();

                return ctx.getStopAction();
            }

            final Buffer payload = httpContent.getContent();
            framePayload = payload.split(payload.position() + PRI_PAYLOAD.length);
        } else {
            framePayload = httpContent.getContent();
        }
```

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2State.java
carryel: 2. As @pzygielo mentioned, it does not need to be declared as volatile.

In contrast, Http2Session's `isPrefaceReceived` is declared as volatile, but this is likely due to multi-thread checks between handleRead() and handleEvent().

It seems that the newly defined Http2State's `isPriReceived` does not need to be declared as volatile. Cases where handleRead() is called simultaneously by multiple threads on the same connection will not occur.

carryel:
I have given you two recommendations regarding the PR you submitted.
Please take another look at this.
Thanks!

11rx4f:
Thanks for the comments.
I’ve addressed the two points you mentioned.

carryel (approved):

PR #2296, opened 2026-05-12
Strict header name and value validation by default

OndroMih:
Also added tests to cover handling corrupted header values. They fail if validation is disabled.

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2025 Contributors to the Eclipse Foundation.
- * Copyright (c) 2010, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2025,2026 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2010,2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -113,13 +113,13 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
      * @see #setRemoveHandledContentEncodingHeaders
      */
     private boolean removeHandledContentEncodingHeaders = false;
-    
+
     public static final String STRICT_HEADER_NAME_VALIDATION_RFC_9110 = "org.glassfish.grizzly.http.STRICT_HEADER_NAME_VALIDATION_RFC_9110";
-    
+
     public static final String STRICT_HEADER_VALUE_VALIDATION_RFC_9110 = "org.glassfish.grizzly.http.STRICT_HEADER_VALUE_VALIDATION_RFC_9110";
-    
+
     private final boolean strictHeaderNameValidation;
-    
+
     private final boolean strictHeaderValueValidation;
 
     /**
@@ -179,7 +179,7 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
 
     /**
      * Callback method, called when {@link HttpPacket} parsing has been completed.
-     * 
+     *
      * @param httpHeader {@link HttpHeader}, which represents parsed HTTP packet header
      * @param ctx processing context.
      *
@@ -205,7 +205,7 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
      * Invoked when either the request line or status line has been parsed.
      *
      * </p>
-     * 
+     *
      * @param httpHeader {@link HttpHeader}, which represents HTTP packet header
      * @param ctx processing context.
      */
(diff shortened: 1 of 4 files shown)


modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
```diff
@@ -815,16 +815,16 @@ protected int parseHeaderName(final HttpHeader httpHeader, final MimeHeaders mim
...
+                    final int eol = checkEOL(parsingState, input, end);
+                    if (eol == 0) { // EOL
+                        // the offset is already increased in the check
+                        return -2;
+                    } else if (eol == -2) { // not enough data
+                        // by keeping the offset unchanged, we will recheck the EOL at the next opportunity.
+                        break;
+                    }
                 }
-            }
```
carryel, line 827: It seems like there is a space in it(Also, the if statement below that is similar to this one.). Would it be possible to edit it?

modules/http/src/test/java/org/glassfish/grizzly/http/ChunkedTransferEncodingTest.java
```diff
@@ -188,12 +210,32 @@ public void testTrailerHeaders() throws Exception {
         headers.put("X-Host", new Pair<>("localhost", "localhost"));
         headers.put("X-Content-length", new Pair<>("2345", "2345"));
 
-        final int packetsNum = 5;
+        final int packetsNum = 1;
```
carryel: I am curious to know what the special reason is for reducing the packet number.

carryel (commented):
LGTM. Please also update the license year in each change file.

modules/http/src/test/java/org/glassfish/grizzly/http/ChunkedTransferEncodingTest.java
OndroMih: I did it while debugging and then forgot to revert it. Thanks for the catch, I'll revert it.

modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
OndroMih: There's no change in the code, just my IDE reformatted it and I didn't notice. I reverted this formatting and force pushed the change.

modules/http/src/test/java/org/glassfish/grizzly/http/ChunkedTransferEncodingTest.java
OndroMih: Reverted and force pushed.

modules/http/src/test/java/org/glassfish/grizzly/http/TestUtils.java
```diff
@@ -0,0 +1,60 @@
...
+                    : String.valueOf(propertyEnabled.booleanValue()));
+        }
+
+        public void unset() {
+            setOrUnsetProperty(previousValue);
+        }
+
+        private void setOrUnsetProperty(String value) {
+            if (value == null) {
+                System.getProperties().remove(propertyName);
```
dmatej: Shouldn't be here rather `System.clearProperty`?

modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
```diff
@@ -454,6 +434,8 @@ private void doHttpRequestTest(Pair<String, String> method, Pair<String, String>
             }
 
             transport.shutdownNow();
+            // wait until the port is unbound
+            Thread.sleep(10);
```
dmatej: Why 10 ms? Is it asynchronous? Is it managed by the operating system or by Grizzly? Is it a leak?

carryel: In the case of the server's unbind, it is treated as blocking.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c36fcdf3f73e7a8000e9d037f4e2abc180fe28ec/modules/grizzly/src/main/java/org/glassfish/grizzly/nio/transport/TCPNIOTransport.java#L266-L274

However, in the case of the client's close in `connection.closeSilently();` from the code immediately above, it is treated as a callback in the NIO environment.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/c36fcdf3f73e7a8000e9d037f4e2abc180fe28ec/modules/grizzly/src/main/java/org/glassfish/grizzly/nio/NIOConnection.java#L528-L540

Ultimately, it seems that a 10ms sleep won't be very meaningful depending on the local environment.

@OndroMih Lastly, would it be possible to apply the improvements mentioned in the @dmatej  comment above?

modules/http/src/test/java/org/glassfish/grizzly/http/TestUtils.java
OndroMih: Changed.

modules/http/src/test/java/org/glassfish/grizzly/http/HttpRequestParseTest.java
OndroMih: I replaced the hardcoded sleep with retrying binding port until it suceeds, with 10 seconds timeout. See the second commit.

carryel (approved):

PR #2298, opened 2026-05-25
Implement HTTP 103 Early Hints (RFC 8297) interim responses

BalusC:
Add `Response.sendEarlyHints()` to send 103 Early Hints interim responses over HTTP/1.1, HTTP/2 and AJP. An interim response carries the headers currently set on the response (e.g. `Link` preload hints), does not commit the response, and may be sent multiple times before the final response. Inspired by Jetty 12 and Tomcat 12.

Prerequisite for a green build of https://github.com/eclipse-ee4j/mojarra/pull/5748.

## Core (`modules/http`)
- New `HttpStatus.EARLY_HINTS_103`.
- `HttpResponsePacket` holds a transient `interimStatus` separate from the final status; serializing an interim response leaves the final status untouched.
- `HttpCodecFilter` serializes the interim status line and headers without marking the headers serialized, so they re-emit with the final response.
- `HttpServerFilter.encodeInitialLine` refactored to share a single private helper with a `(packet, status)` overload — preserves `CustomReasonPhrase` on the final response only.
- 100-Continue and 1xx interim responses now share a single `interimStatus` mechanism (removed `setAcknowledgement`/`isAcknowledgement`/`acknowledged` from `HttpResponsePacket`; the four parallel `isAcknowledgement` branches in HTTP/1.1, AJP and HTTP/2 codecs collapse into the interim branch). `OutputBuffer.acknowledge()` and `writeInterimResponse()` collapse into the latter.
- `HttpServerFilter` no longer runs `prepareResponse` (which injects `Date`/`Content-Type`/`Content-Length`) on interim packets — those mutations belong to the final response.

## HTTP/2 (`modules/http2`)
- `DefaultOutputSink` emits the interim as a HEADERS frame (END_HEADERS, no END_STREAM) without committing the stream.
- `Http2ClientFilter` accepts inbound interim 1xx responses instead of failing with `PROTOCOL_ERROR`.
- `DecoderUtils.finalizeKnownHeader` had a missing `return` on `case "expect":` (and `case "te":`) causing every HTTP/2 `Expect: 100-continue` request to be rejected at decode time — fixed. HTTP/2 100-Continue now works end-to-end via the unified interim mechanism.
- `HttpServerFilter#prepareRequest` uses the new `Protocol#isAtLeast(HTTP_1_1)` for the legacy-protocol check.

## AJP (`modules/http-ajp`)
- `AjpHandlerFilter` and `AjpMessageUtils` encode interim responses as `SEND_HEADERS` packets without committing — the existing 100-Continue mechanism generalised. Tomcat NO-OPs early hints over AJP; we implement because the AJP13 protocol does not forbid multiple `SEND_HEADERS` packets, and forwarding is the upstream connector's concern.

## Servlet (`modules/http-servlet`)
- `HttpServletResponseImpl.sendEarlyHints()` (Servlet 6.2). The Servlet 6.2 spec signature does not declare `IOException`; the (effectively unreachable) async write path's checked exception is logged and swallowed — matches glassfish web-core's `ResponseFacade`.
- `ServletHandler.AckActionImpl.acknowledge()` no longer leaks status `100` onto the final response after the unification (regression test added).
- `jakarta.servlet-api` bumped 6.1.0 → 6.2.0-M2.

## Tests
- `EarlyHintsTest` (HTTP/1.1), `Http2EarlyHintsTest`, `AjpEarlyHintsTest` — interim 103 carries `Link`, final response still has it.
- `Http2ContinueTest` — HTTP/2 100-Continue end-to-end.
- `ServletHttpContinueTest` regression for the auto-ack leak.
- `HttpContinueTest`, `BasicAjpTest.test100ContinuePost`, `ServletHttpContinueTest` all still green after unification.

### DIFF
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
@@ -179,6 +179,16 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
     abstract Buffer encodeInitialLine(HttpPacket httpPacket, Buffer output, MemoryManager memoryManager);
 
     /**
+     * Variant of {@link #encodeInitialLine(HttpPacket, Buffer, MemoryManager)} that overrides the status used in the
+     * status line, for serializing interim (1xx) responses whose status differs from the final response status. The
+     * default implementation ignores {@code status} and delegates to the single-arg form; only the server-side filter
+     * needs to honour the override.
+     */
+    Buffer encodeInitialLine(HttpPacket httpPacket, HttpStatus status, Buffer output, MemoryManager memoryManager) {
+        return encodeInitialLine(httpPacket, output, memoryManager);
+    }
+
+    /**
      * Callback method, called when {@link HttpPacket} parsing has been completed.
      *
      * @param httpHeader {@link HttpHeader}, which represents parsed HTTP packet header
@@ -1514,13 +1524,8 @@ public abstract class HttpCodecFilter extends HttpBaseFilter implements Monitori
                 if (response.isInterimResponse()) {
                     // Interim (1xx) responses are only emitted for HTTP/1.1; the caller guards on the request protocol,
                     // so the response protocol is HTTP/1.1 here.
-                    final HttpStatus interimStatus = response.getInterimStatus();
                     encodedBuffer = memoryManager.allocateAtLeast(2048);
-                    encodedBuffer = put(memoryManager, encodedBuffer, response.getProtocol().getProtocolBytes());
-                    encodedBuffer = put(memoryManager, encodedBuffer, Constants.SP);
-                    encodedBuffer = put(memoryManager, encodedBuffer, interimStatus.getStatusBytes());
-                    encodedBuffer = put(memoryManager, encodedBuffer, Constants.SP);
-                    encodedBuffer = put(memoryManager, encodedBuffer, interimStatus.getReasonPhraseBytes());
+                    encodedBuffer = encodeInitialLine(httpHeader, response.getInterimStatus(), encodedBuffer, memoryManager);
                     encodedBuffer = put(memoryManager, encodedBuffer, CRLF_BYTES);
                     onInitialLineEncoded(httpHeader, ctx);
 
--- a/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
+++ b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
@@ -44,6 +44,7 @@ import org.glassfish.grizzly.http.util.Constants;
 import org.glassfish.grizzly.http.util.CookieHeaderParser;
 import org.glassfish.grizzly.http.util.DataChunk;
 import org.glassfish.grizzly.http.util.Header;
+import org.glassfish.grizzly.http.util.HttpStatus;
 import org.glassfish.grizzly.http.util.MimeHeaders;
 import org.glassfish.grizzly.memory.Buffers;
 import org.glassfish.grizzly.memory.CompositeBuffer;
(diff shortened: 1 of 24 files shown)


modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
```diff
@@ -1510,6 +1511,33 @@ protected final Buffer encodeHttpPacket(final FilterChainContext ctx, final Http
...
+                if (response.isInterimResponse()) {
+                    // Interim (1xx) responses are only emitted for HTTP/1.1; the caller guards on the request protocol,
+                    // so the response protocol is HTTP/1.1 here.
+                    final HttpStatus interimStatus = response.getInterimStatus();
+                    encodedBuffer = memoryManager.allocateAtLeast(2048);
+                    encodedBuffer = put(memoryManager, encodedBuffer, response.getProtocol().getProtocolBytes());
+                    encodedBuffer = put(memoryManager, encodedBuffer, Constants.SP);
+                    encodedBuffer = put(memoryManager, encodedBuffer, interimStatus.getStatusBytes());
+                    encodedBuffer = put(memoryManager, encodedBuffer, Constants.SP);
+                    encodedBuffer = put(memoryManager, encodedBuffer, interimStatus.getReasonPhraseBytes());
```
carryel: How about defining a private method with an added parameter for HttpStatus to enable code reuse and continue supporting CustomReasonPhrase as before?

For example, it is as follows:

```
encodedBuffer = encodeInitialLine(httpHeader, interimStatus, encodedBuffer, memoryManager);

...

@Override
Buffer encodeInitialLine(HttpPacket httpPacket, Buffer output, MemoryManager memoryManager)
    return encodeInitialLine(httpPacket, ((HttpResponsePacket) httpPacket).getHttpStatus(), output, memoryManager);
}

private Buffer encodeInitialLine(HttpPacket httpPacket, HttpStatus status, Buffer output, MemoryManager memoryManager) {
    ...
}
```

carryel:
Although Servlet spec 6.2 has not been released yet, how about adding it to `HttpServletResponseImpl` for the future?

Reference) https://github.com/jakartaee/servlet/pull/672/files

For example, it is as follows:

```
    /**
     * Sends a 103 response to the client using the current response headers. This method does not commit the response and
     * may be called multiple times before the response is committed. The current response headers may include some headers
     * that have been added automatcially by the container.
     * <p>
     * This method has no effect if called after the response has been committed.
     *
     * @since Servlet 6.2
     */
    public void sendEarlyHints() {
        if (response == null) {
            throw new IllegalStateException("Null response object");
        }
        return response.sendEarlyHints();
    }
```

modules/http/src/main/java/org/glassfish/grizzly/http/io/OutputBuffer.java
```diff
@@ -376,6 +376,19 @@ public void acknowledge() throws IOException {
...
+     * This does not commit the response.
+     *
+     * @throws IOException if an error occurs writing the interim response.
+     */
+    public void writeInterimResponse() throws IOException {
+
+        ctx.write(outputHeader, IS_BLOCKING);
+
+    }
+
```
carryel, line 378: It appears that `public void acknowledge()` could be replaced with that part. It would be fine to clean it up now, but since there seem to be some parts of the acknowledge section—specifically the HttpStatus 100 continue logic—that can be combined with the Interim Response to be added this time, it might be possible to clean it up further.

carryel:
I think the AJP Protocol also needs to be modified. It would be good if you referred to the section below and considered it in a similar manner.

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/e5c322debf631342d8eea888abf898884bfa98a6/modules/http-ajp/src/main/java/org/glassfish/grizzly/http/ajp/AjpHandlerFilter.java#L213-L218

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/e5c322debf631342d8eea888abf898884bfa98a6/modules/http-ajp/src/main/java/org/glassfish/grizzly/http/ajp/AjpMessageUtils.java#L385-L389

BalusC:
Great feedback. Let me take a look.

BalusC:
AJP is going to be more work than anticipated because there is no foundation and it's even not implemented in Tomcat so it cannot be used as reference.

BalusC:
@carryel: all your feedback has been addressed. PR description is also updated.

modules/http/src/main/java/org/glassfish/grizzly/http/HttpCodecFilter.java
BalusC: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2298/commits/f1a0e9e9b12f77f44dead0c9bfd9dc0167da75d2

modules/http/src/main/java/org/glassfish/grizzly/http/io/OutputBuffer.java
BalusC: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2298/commits/003d930a025f5f31dc3e5440657469285e666052

carryel (approved):

pom.xml
```diff
@@ -91,7 +91,7 @@
     <properties>
         <java.version>21</java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <servlet-version>6.1.0</servlet-version>
+        <servlet-version>6.2.0-M2</servlet-version>
```
carryel: Finally, I have one more concern. GlassFish 8.x supports Jakarta EE 11 and Jakarta Servlet 6.1.
Grizzly 5.0.x supports Jakarta EE 11 and Java 21+.
I would like to hear your and others' opinions on whether we should switch the current main branch to Jakarta Servlet 6.2.0-M2.

BalusC: Then we have 2 options:
- Bump to Grizzly 6.0.0-SNAPSHOT (and adjust GlassFish PR)
- OR revert servlet version bump and remove `@Override` from `HttpServletResponseImpl` (we'll add back later when on real Servlet 6.2)

carryel: Since Grizzly 5.x was released not long ago and it is impossible to predict when 6.x will be released, I think the latter method(Servlet 6.1) would be better to consider potential impacts.

BalusC: Ok, done: https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2298/commits/bd6337db5c8420f6139a413b6beb6a6c579cbf3f

BalusC:
Thank you for merging. What's the approximate release date for 5.0.2?

carryel:
> What's the approximate release date for 5.0.2?

As far as I know, Grizzly does not release updates regularly.

I believe that if there is a need or if a sufficient number of features are gathered, it will be possible to release updates at any time upon request.

OndroMih:
We will release Grizzly 5.0.2 soon, there's also an important fix pending

PR #2299, opened 2026-06-02
Fixed "Keep-alive connection processing breaks with SameThreadIOStrategy after suspend/resume"

### LINKED ISSUE #2297: Keep-alive connection processing breaks with SameThreadIOStrategy after suspend/resume
### The issue

When a `HttpHandler` suspends the response on the selector thread and resumes it asynchronously from a worker, the IO-event interest bookkeeping in [`SameThreadIOStrategy.InterestLifeCycleListenerWhenIoEnabled`](https://github.com/eclipse-ee4j/glassfish-grizzly/blob/main/modules/grizzly/src/main/java/org/glassfish/grizzly/strategies/SameThreadIOStrategy.java#L97) ends up with OP_READ disabled on the connection but never re-enabled. The selector then parks in `select()` and the next pipelined request on the keep-alive connection is never picked up - the client times out.

### Reproducer test

I have created reproducer test for this issue in my fork https://github.com/pavelhoral/glassfish-grizzly/commit/f9735764c02b014946196c0123416ca68a9d500b . Action run (and test failure) can be seen in https://github.com/pavelhoral/glassfish-grizzly/actions/runs/26229540441/job/77185965304#step:5:1634 .

### Real world use-case

I am using `SameThreadIOStrategy` and delegating to own worker because I need the processing to take place in pre-existing thread pool that is unrelated to Grizzly.

### How to fix?

I am not exactly sure where a proper fix should be implemented here (is the issue with `wasSuspended` context flag or with the IO strategy?). That is why I am creating this report and not a PR with some naive fix. I can try to implement a proper fix if you give mi some advice / directions.

carryel:
- Issue #2297
- When `suspend()` is called on the `Response`, the actual suspend of the corresponding context is performed so that `OP_READ` can be enabled even under `SameThreadIOStrategy`.
- Added test cases to verify normal operation across various IOStrategy.

### DIFF
--- a/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
+++ b/modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -166,7 +166,7 @@ public class FilterChainContext implements AttributeStorage {
             @Override
             public void run() {
                 try {
-                    if (state == State.SUSPEND) {
+                    if (isSuspended()) {
                         state = State.RUNNING;
                     }
 
@@ -214,7 +214,7 @@ public class FilterChainContext implements AttributeStorage {
     public void resume(final NextAction nextAction) {
         internalContext.resume();
         try {
-            if (state == State.SUSPEND) {
+            if (isSuspended()) {
                 state = State.RUNNING;
             }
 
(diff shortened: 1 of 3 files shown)


modules/grizzly/src/main/java/org/glassfish/grizzly/filterchain/FilterChainContext.java
```diff
@@ -262,6 +262,15 @@ public State state() {
         return state;
     }
 
+    /**
+     * Return <tt>true<//tt> if that {@link FilterChainContext#suspend()} has been invoked
+     *
+     * @return <tt>true<//tt> if that {@link FilterChainContext#suspend()} has been invoked
+     */
+    public boolean isSuspended() {
```
pzygielo, line 265: If I'd decide to have such comment (which with the name of the method and its location I find unneeded) - I'd keep only @return and not repeat the same in introduction. And {@link #suspend} to keep it local.

Then - I'd use {@code true} instead (it would also save me for free from bad closing of tt element).

> `suspend` has been invoked

But what it `resume` was also invoked?

I suggest to drop this javadoc.

carryel: Removed unnecessary and ambiguous javadoc.

dmatej (approved):

PR #2303, opened 2026-08-18
Initialize internal variables when recycling objects

carryel:
+ When recycling, uninitialized variables may lead to unintended behavior in HTTP/2.
+ This issue is related to `work_items/219` on https://gitlab.eclipse.org, which was reported privately ("Repeated HTTP/2 requests result in frequent HTTP 431 errors and...").


carryel:
@OndroMih @pzygielo Could you possibly review the changes?

The fix is ​​simple. I have improved the part where initialization was missing during recycle in the place where variables were defined. Also, since some variables were often initialized individually within the recycle methods of their subclasses rather than where they were defined, I moved this part to the defined class to prevent future errors.

I confirmed that the issue occurred after deploying the war to Glassfish, and I also confirmed that there are no issues after applying the patch.

pzygielo (approved):

Issue #2304, opened 2026-08-20
Off-by-one Error leads to Premature HTTP/2 Stream Refusal


mkarg:
# Problem Description

When configuring HTTP/2 max stream count to `n`, Grizzly correctly reports `SETTINGS_MAX_CONCURRENT_STREAMS` with a value of `n` in `SETTINGS` (proven using `nghttp -v`), but answers with `REFUSED_STREAM` when actually using the `n`ᵗʰ stream, while the `1`ˢᵗ up to the `n-1`ᵗʰ stream work well (proven using `h2load -m <n>` -which fails- and `h2load -m <n-1>` -which works- using a synthetically slowed-down responder to prevent premature end-of-stream until the stream capacity is actually saturated).

# Effect

Some clients are unable to deal with the situation that server is sending `SETTINGS_MAX_CONCURRENT_STREAMS` = `n`, but `REFUSED_STREAM` for stream `n` already. The negative effect is different and specific to the client:
* `h2load` fails to send *any* requests on *any* stream!
* Firefox successfully uses the opened `1`ˢᵗ to `n-1`ᵗʰ streams, but reports an *"unknown error"* to the JavaScript application which tried to fetch. As the application is not told the actual cause, the requested fetch on the `n`ᵗʰ stream fails with "**unknown** error"! Typically JavaScript applications are unable to deal with that (in the worst case) or try to send again (which leads to performance penalties in the best case, or which leads to *duplicate execution* in case of *other* "unkown" errors besides the off-by-one problem - as the application cannot *tell the difference*).

# Cause

Apparently *somewhere* in counting the number of open streams there must be an [Off-by-one error](https://wikipedia.org/wiki/Off-by-one_error). Unfortunately I could no see one, as the code *looks* correct:

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/923775369872e875fa078b144cd879c330882bb6/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Session.java#L944-L948

https://github.com/eclipse-ee4j/glassfish-grizzly/blob/923775369872e875fa078b144cd879c330882bb6/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Session.java#L993-L995

Nevertheless, as the described effect is *proven and reproducible* (using Payara 7.2026.7, which contains Grizzly 5.0.2), the count *is* off-by-one frequently, which means that there **must** be a counting error *somewhere*. 🥴 

# Workaround

[...]

PR #2305, opened 2026-08-20
Workaround for off-by-one error in HTTP/2 max concurrent streams checks

mkarg:
This PR proposes an *acceptable workaround* for the issue reported and described in #2304 to mitigate the effect unless a solution is found for the *root* cause.

### DIFF
--- a/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Session.java
+++ b/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Session.java
@@ -941,7 +941,7 @@ public class Http2Session {
                 return null; // if the session is closed is set - return null to ignore stream creation
             }
 
-            if (concurrentStreamsCount.get() >= getLocalMaxConcurrentStreams()) {
+            if (concurrentStreamsCount.get() > getLocalMaxConcurrentStreams()) {
                 // throw Session level exception because headers were not decompressed,
                 // so compression context is lost
                 throw new Http2SessionException(ErrorCode.REFUSED_STREAM);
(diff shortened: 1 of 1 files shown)


mkarg:
Once #2309 is merged, I will revoke this draft proposal, as we have a real solution then (so I hope).

pzygielo:
@mkarg - have you had a chance to verify that #2309 helps in your full case?

mkarg:
> @mkarg - have you had a chance to verify that #2309 helps in your full case?

Not yet. I first need to find out if I simply can build Grizzly on my own and put it into Payara 7 (for some JARs that is not working, as Payara comes with private copies).

pzygielo:
> Not yet. I first need to find out if I simply can build Grizzly on my own and put it into Payara 7 (for some JARs that is not working, as Payara comes with private copies).

Here is what I was doing initially to patch Payara after building grizzly locally:

```
GRJF=/tmp/payara7/glassfish/modules/nucleus-grizzly-all.jar 

cd modules/http2/target/classes

jar vuf ${GRJF} org/glassfish/grizzly/http2/Http2Session*.class
```

(now with new PR this would need more files to be updated of course).

pzygielo:
```
git checkout 0d47efb # [maven-release-plugin] prepare for next development iteration
git cherry-pick bbaef6b # 2309/head^
git cherry-pick 69e0f15 # 2309/head
```
to build just 5.0.2 patched

and then

```
mvn -DskipTests -am -pl modules/http2 clean package
```
to build just patched module

Will check with the reproducer from issue a bit later.

mkarg:
Thank you @pzygielo for this detailed instructions. Using them I was able to patch Payara 7.2026.8 and run the reproducer again. 🙂 

**I am very happy to confirm that actually the off-by-one error is *resolved* with #2309. 🚀**

Even a longer 5 minutes run using `-D 300` confirmed that now zero HTTP/2 stream errors happen anymore. 🤩

PR #2307, opened 2026-09-09
Fix premature SafeFutureImpl completion detection

### LINKED ISSUE #2306: SafeFutureImpl completion race: RESULT state is treated as terminal, allowing get() to return null
`SafeFutureImpl.Sync.ranOrCancelled()` treats the intermediate RESULT state as terminal. A concurrent `get()` can consequently return before the result or exception is published, including returning null when the producer is completing the future with a non-null connection.

### Affected component and versions

Component: `grizzly-framework`, `org.glassfish.grizzly.impl.SafeFutureImpl`.

I checked the published source JARs for 4.0.2, 4.1.0-M1, 5.0.0, 5.0.1, and 5.0.2; all contain the same predicate and state values. The current `main` source also contains them:
https://github.com/eclipse-ee4j/glassfish-grizzly/blob/main/modules/grizzly/src/main/java/org/glassfish/grizzly/impl/SafeFutureImpl.java

### Failure mechanism

The states are READY=0, RESULT=1 (result/exception being written), RAN=2, and CANCELLED=3. The completion predicate is:

```java
private boolean ranOrCancelled() {
    return (getState() & (RAN | CANCELLED)) != 0;
}
```

Because `(1 & (2 | 3)) != 0`, this returns true while the producer is still in RESULT.

A permitted interleaving is:

1. Producer enters `innerSet(nonNullConnection)` and successfully changes READY to RESULT.
2. Producer is descheduled before `result = v`.
3. Consumer calls `get()` or timed `get()`. `tryAcquireShared()` accepts RESULT as complete, and `innerGet()` returns the still-null result.
4. Producer stores the non-null connection and publishes RAN.

There is an analogous window in `innerSetException()`: the consumer can return null before the exception is stored instead of throwing `ExecutionException`. `isDone()` also reports true prematurely.

### Observed downstream symptom

With Grizzly 4.0.2, oncrpc4j-core 3.4.2, and Java 21, a UDP connection future used for local portmap publication returned null and the caller failed in `GrizzlyRpcTransport` while dereferencing `connection.getPeerAddress()`.

The source race above is a concrete explanation consistent with that failure. I have not captured the exact production thread interleaving or run a standalone concurrency reproducer, so I am distinguishing the source defect from proof of that particular occurrence.

### Suggested correction

Read the state once and accept only terminal states:

```java
private boolean ranOrCancelled() {
    int state = getState();
    return state == RAN || state == CANCELLED;
}
```

Keep the intermediate RESULT state: it arbitrates completion/cancellation while the producer writes the result or exception. Readers must wait for the subsequent RAN publication rather than treating RESULT as completion.

### Suggested regression coverage

Exercise concurrent non-null result publication and exception publication against both `get()` variants and `isDone()`. A deterministic test that pauses completion after READY→RESULT should verify that readers remain blocked until RAN; then they must receive the non-null result or the expected `ExecutionException`. Verify cancellation behavior as well.

jcalcote:
`SafeFutureImpl` can report completion and let `get()` return before the producer has published its result or exception. The completion predicate uses a bitmask even though the state values overlap: `RESULT=1` passes the mask formed from `RAN=2` and `CANCELLED=3`.

Read the state once and explicitly accept only `RAN` or `CANCELLED`. This preserves the intermediate publication state and all existing state transitions.

Fixes #2306.

The regression tests hold the synchronizer in `RESULT` using test-only reflection/method handles, without adding production hooks or JVM module-opening flags. They check `isDone()`, timed retrieval, cancellation arbitration, deferred callbacks, and queued readers for both `get()` variants. The fixture then restores `READY` and invokes the normal result/exception completion path to verify publication and reader release. Cancellation is covered separately.

Validation on Java 21 and Maven 3.9.9:

- Before the fix: four publication tests fail; the cancellation test passes.
- After the fix: all five new tests pass.
- `mvn -B -ntp -pl modules/grizzly -am verify`: successful; 388 tests, 0 failures, 0 errors, 10 skipped.
- `git diff --check`: clean.

A backport to `4.1.x` would be appreciated if feasible, for inclusion in a future release of that line. I can provide a separate backport PR if helpful. Otherwise, our application can retain its workaround until it moves to a 5.x release containing this fix.


modules/grizzly/src/main/java/org/glassfish/grizzly/impl/SafeFutureImpl.java
```diff
@@ -285,7 +285,8 @@ private final class Sync extends AbstractQueuedSynchronizer {
         private Throwable exception;
 
         private boolean ranOrCancelled() {
-            return (getState() & (RAN | CANCELLED)) != 0;
+            int state = getState();
```
carryel: Even though it doesn't affect functionality, could we add `final` for the sake of consistency with the rest of the code?

```
final int state = getState();
```

carryel: It would also be good to include the license information in the header.

```
/*
 * Copyright (c) 2026 Contributors to the Eclipse Foundation.
 * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
...
```

carryel:
> Could this fix also be backported to the 4.0.x branch and included in a future 4.0.x maintenance release? 

I took a look at the branch structure.
Unfortunately, a 4.0.x release does not seem likely.
While a 4.1.0 release might be possible, nothing is guaranteed.
The difference between 4.0.x and 4.1.x (which has not yet had a release) is presumably, at a minimum, the required Java version.

```
4.1.x : This is the sustaining branch for 4.1.x (Jakarta EE 11 with Java 17 support). (no release yet)
4.0.x : This is the sustaining branch for 4.0.x (Jakarta EE 10). (latest release is 4.0.2)
```

modules/grizzly/src/main/java/org/glassfish/grizzly/impl/SafeFutureImpl.java
jcalcote: Thanks for the review. Added `final` to the local state variable and the 2026 Eclipse contributor copyright line, preserving the existing Oracle copyright and license text. Pushed both changes; the framework build and all five SafeFutureImpl regression tests pass on Java 21.

jcalcote:
Thank you for checking the release options. Would a backport to `4.1.x` be possible, for inclusion if a 4.1.x release is made? We can evaluate that line for our application, and I would be happy to provide a separate backport PR if helpful.

If that is not feasible, no problem: we can retain our workaround until we move to a 5.x release containing this fix, then reconsider enabling RPC advertisement. We appreciate your help getting the underlying race fixed upstream.

carryel (approved):

carryel:
I have also applied the changes to the 4.1.x branch.
I will check if a 4.x release is feasible alongside the upcoming Grizzly 5.x release.
I will share an update once the release process moves forward.

jcalcote:
Thank you! :)

PR #2308, opened 2026-09-10
Fixed for HttpUpgradeTest that occasionally caused failures

carryel:
- Since blocking and non-blocking reads can occur simultaneously, the blocking read is removed.
- Details: 
  - https://jakarta.ee/specifications/servlet/6.1/jakarta-servlet-spec-6.1#upgrade-processing `The Application Developer has the responsibility for thread safe access to the ServletInputStream and ServletOutputStream while processing HTTP upgrade.`
  - https://ci.eclipse.org/glassfish/view/Grizzly/job/Grizzly-master_nightly/org.glassfish.grizzly$grizzly-http-servlet/3135/testReport/junit/
```
expected:<Hell[]oWorld> but was:<Hell[l]oWorld>

Stack Trace

junit.framework.ComparisonFailure: expected:<Hell[]oWorld> but was:<Hell[l]oWorld>
	at junit.framework.Assert.assertEquals(Assert.java:100)
	at junit.framework.Assert.assertEquals(Assert.java:107)
	at junit.framework.TestCase.assertEquals(TestCase.java:260)
	at org.glassfish.grizzly.servlet.httpupgrade.HttpUpgradeTest.testEchoProtocolUpgrade(HttpUpgradeTest.java:131)
```


carryel:
This modification to the test case is related to the following issue:
https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2281

Previously, `isReady()` was always required; however, this requirement has been changed, and the resulting modification increased the likelihood of test case errors.

Since `ServletInputStream` is not potentially thread-safe, performing `read()` operations across multiple threads in the original test case would have required the addition of synchronization. Instead of implementing thread-safe access, the test case was modified to avoid multi-threaded access altogether.

pzygielo (approved):

PR #2309, opened 2026-09-10
Fix HTTP/2 concurrent-stream limit: refuse only the stream, stop counting it before END_STREAM is sent

### LINKED ISSUE #2304: Off-by-one Error leads to Premature HTTP/2 Stream Refusal

renatsaf:
Two defects in the handling of the local `SETTINGS_MAX_CONCURRENT_STREAMS` limit, both behind #2304.

### 1. A stream over the limit terminated the whole session

`Http2Session.acceptStream()` answered a HEADERS frame opening a stream over the limit with `Http2SessionException(REFUSED_STREAM)`: GOAWAY, connection close, and every stream in progress on the connection failed. The check ran before the header block was decoded, and an undecoded block leaves the HPACK dynamic table out of sync, hence the session error. RFC 9113, section 5.1.2 requires a stream error. Browsers report it as `ERR_HTTP2_PROTOCOL_ERROR` on one request and `ERR_CONNECTION_CLOSED` on the requests next to it (eclipse-ee4j/glassfish#25954).

- `acceptStream()` throws `Http2StreamException(REFUSED_STREAM)`, so only that stream is reset. The stream ID validity checks run first, and the refused ID is recorded as used, so frames still arriving for it are handled as for a closed stream.
- `DecoderUtils.skipHeaders()` runs the refused header block through the HPACK decoder; called from `Http2ServerFilter` and from the client `PUSH_PROMISE` path.

### 2. A stream kept counting after its END_STREAM frame was written

The count was lowered in `deregisterStream()`, which runs after the END_STREAM frame has been handed to the connection (`DefaultOutputSink.flushToConnectionOutputSink()` writes the frame, then calls `terminate()`). A peer opening a new stream as soon as the previous one completed could be refused although it never exceeded the limit. With the limit set to 1 and 5000 strictly sequential requests on one connection, request 608 was refused.

- `Http2Session.onSendEndOfStream()` is called right before a frame with END_STREAM is written. If the stream input is already closed, the stream is marked closing and no longer counts toward the limit.
- Deregistration, and closing the session once no stream is left after GOAWAY, still happen after the write.

### Relation to #2305

#2305 replaces `>=` with `>`: one stream more than advertised is accepted, and a stream over that limit still terminates the session. This change keeps the advertised limit exact and addresses both causes.

### Tests

`Http2MaxConcurrentStreamsTest`, limit set to 1:

- `testStreamOverLimitIsRefusedWithoutClosingConnection` fails without the first commit (connection closed). It also checks that the request following the refused one decodes a header indexed from the refused request's header block.
- `testStreamOpenedAfterPreviousOneCompletedIsNotRefused` sends 5000 sequential requests; fails without the second commit (`request 485 got no response, its stream was reset`).

`mvn -pl modules/http2 -am test`: http2 module 266 tests, 0 failures.

- Fixes #2304 and closes #2305.

https://claude.ai/code/session_01HFHkqh3iohF6ZFXWehA7cP


pzygielo:
- I like that #2305 was also considered, analyzed and referenced here.

mkarg:
I patched PR #2309 into Payara 7.2026.8 locally and performed successfully a five minutes test-drive with my reproducer from issue #2305 using `h2load`. It produced **zero** HTTP/2 stream faults. 😍 

**I am very happy to confirm that actually the off-by-one error is *resolved* with #2309. 🚀**

mkarg:
Tonight successfully performed three-hours load test using `h2 -D 10800`: Zero errors. 😍

mkarg:
@eclipse-ee4j/ee4j-grizzly-committers Kindly asking for review! 🙏

dmatej:
> @eclipse-ee4j/ee4j-grizzly-committers Kindly asking for review! 🙏

It is too much complexity for my brain, but it looks good. For sure, before the release, I will probably create some "snapshot-based" glassfish branch and try to pass tests with dependencies from Maven Central Snapshots. Which means this would have to be merged. Another option is to build grizzly locally and test everything locally - but it seems you already made this successfully.

dmatej (approved):

PR #2310, opened 2026-09-12
Use a monotonic clock for HTTP session expiration

### LINKED ISSUE #2175: Session timeout does not tolerate system time changes
The `Session` object has built-in handling for timeouts, but it uses `System.currentTimeMillis()` as its source of time. This is affected by operating system time changes, which has a number of strange/problematic effects:

1. A time change forward shortens or immediately invalidates the session
2. A time change backward lengthens the session or makes it effectively perpetual

In our application we are using the session timeout to invalidate the user's authentication, so this behavior has some pretty undesirable effects and possible security implications.

Devices that are connected to an NTP server may be affected by this behavior shortly after startup, when the time resynchronizes with the server. On some devices it may also be possible for an unprivileged user to modify the system time as a way to circumvent authentication time limits on the server.

Would it be possible to use `System.nanoTime()` as a monotonic time source for the session expiration calculation?

# Scenarios

The scenarios below assume the following request handling code:
```java
public void service(Request request, Response response) throws Exception {
    final Session session = request.getSession();

    if (session.isNew()) {
        System.out.println("New session used");
        session.setSessionTimeout(TIMEOUT_MILLIS);
    } else {
        System.out.println("Old session used");
    }
}
```

## Time moves backward

**Scenario**

```java
final long TIMEOUT_MILLIS = 10_000; // 10 seconds.
```

1. Run server using the code above
1. Send first request
1. Change system time by `-2` minutes
1. Wait 15 seconds
1. Send second request

**Expected**

Output should be:
```
New session used
New session used
```

No request is made to the server within the 10 second timeout. The session should expire, and the next request should trigger the creation of a new session.

**Actual**

Output is:
```
New session used
Old session used
```

The expiry calculation uses the underlying system time (which was moved backwards), so the session is not yet expired even though more than 10 seconds have passed.

## Time moves forward

**Scenario**

```java
final long TIMEOUT_MILLIS = 60_000; // 1 minute.
```

1. Run server using the code above
1. Send first request
1. Change system time by `+2` minutes
1. Send second request

**Expected**

Output should be:
```
New session used
Old session used
```

A request is made to the server within the 1 minute timeout. The session should still be valid, and should be reused.

**Actual**

Output is:
```
New session used
New session used
```

The expiry calculation uses the underlying system time (which was moved forwards), so the session is immediately expired. A new session is created as a result.

renatsaf:
Fixes #2175

### Problem

`Session` records the last access with `System.currentTimeMillis()`, and the `DefaultSessionManager` sweeper expires a session when `currentTimeMillis() - getTimestamp() > getSessionTimeout()`. A system clock change (NTP step after boot, VM resume, manual change) therefore alters session lifetime:

- clock moved forward: sessions expire immediately
- clock moved back: sessions outlive their timeout by the size of the step

In addition, `DefaultSessionManager.getSession()` only checked `isValid()`, so a session that had already exceeded its timeout was still returned until the next sweep, up to 5 s later.

### Changes

- `Session`: new monotonic `lastAccessedNanos`, set in the constructor and in `access()`, and `isExpired(long nowNanos)`. `getTimestamp()` and `getCreationTime()` keep returning epoch milliseconds, as required by `HttpSessionImpl.getLastAccessedTime()` / `getCreationTime()` in `http-servlet`.
- `Session.setTimestamp(long)` keeps affecting expiry: the monotonic access time is moved back by the same amount. A timestamp in the future is treated as the current time, so the session still expires at most one timeout from the call, and the shift is bounded, so extreme values (`-1`, `Long.MIN_VALUE`) cannot overflow into a session that never expires. `getTimestamp()` returns the value as given.
- `Session`: fields read by the sweeper thread are `volatile`.
- `DefaultSessionManager`: the sweeper uses `isExpired(System.nanoTime())`; `getSession()` invalidates and removes a session that expired since the last sweep.

On Linux `System.nanoTime()` is based on `CLOCK_MONOTONIC`, which does not advance while the host is suspended, so time spent in suspend does not count as idle time.

### Tests

New `SessionExpirationTest` (7 tests, no sleeps or clock changes): expiry boundary, `access()` restarting idle time, no timeout, `setTimestamp` shift including future and extreme values, and `DefaultSessionManager.getSession()` not returning a lapsed session before the sweep.

`mvn -pl modules/http-server,modules/http-servlet -am test -Dtest='SessionExpirationTest,HttpSessionTest'` on JDK 25 (Temurin 25.0.3+9): http-server `HttpSessionTest` 7/7, `SessionExpirationTest` 7/7, http-servlet `HttpSessionTest` 3/3. Full module suites were not run locally.


modules/http-server/src/main/java/org/glassfish/grizzly/http/server/Session.java
```diff
@@ -39,12 +49,12 @@ public class Session {
...
      * Is this Session valid.
      */
-    private boolean isValid = true;
+    private volatile boolean isValid = true;
 
     /**
      * Is this session new.
      */
-    private boolean isNew = true;
+    private volatile boolean isNew = true;
```
carryel: Do you think the `isNew` variable also needs to be `volatile`?
The `isNew` variable is accessed only via `HttpSession#isNew()`.

modules/http-server/src/main/java/org/glassfish/grizzly/http/server/Session.java
```diff
@@ -54,12 +64,18 @@ public class Session {
...
      * Timeout
      */
-    private long sessionTimeout = -1;
+    private volatile long sessionTimeout = -1;
 
     /**
      * Creation time stamp.
      */
-    private long timestamp = -1;
+    private volatile long timestamp = -1;
```
carryel: Do you think the `timestamp` variable also needs to be `volatile`?
The `timestamp` variable is accessed only via `HttpSession#getLastAccessedTime()`.

modules/http-server/src/main/java/org/glassfish/grizzly/http/server/Session.java
renatsaf: No, reverted in 41444b0. `isNew` is set by `access()` in `Request.doGetSession` and read by `HttpSession#isNew()` on the same request thread, so it has no cross-thread reader.

modules/http-server/src/main/java/org/glassfish/grizzly/http/server/Session.java
renatsaf: No, reverted in 41444b0. Before this change the only cross-thread reader of `timestamp` was the sweeper; it now uses `lastAccessedNanos`, so `timestamp` is read only by `getLastAccessedTime()`, which checks `isValid` first.

`volatile` stays on:
- `lastAccessedNanos` and `sessionTimeout`: read by the sweeper. `createSession` publishes the session through `putIfAbsent` before `Request.doGetSession` calls `setSessionTimeout`.
- `isValid`: read by the sweeper and by `getSession` on other request threads.

modules/http-server/src/main/java/org/glassfish/grizzly/http/server/Session.java
```diff
@@ -198,11 +215,16 @@ public long getTimestamp() {
...
     /**
-     * Set the timestamp when this session was accessed the last time.
+     * Set the timestamp when this session was accessed the last time. The idle time used for expiration is shifted by the
+     * same amount.
      * 
      * @param timestamp a long representing when the session was accessed the last time
      */
     public void setTimestamp(long timestamp) {
+        final long nowMillis = System.currentTimeMillis();
+        final long boundedTimestamp = Math.max(nowMillis - MAX_TIMESTAMP_SHIFT_MILLIS,
```
carryel: I spent quite a bit of time thinking about this specific aspect.
What would happen if the `timestamp` provided is in the future rather than the present? For instance, suppose the `timestamp` is set to a time five minutes ahead of the current time (perhaps due to a system clock error).
Even with a 30-minute session timeout, it seems the session would effectively expire 35 minutes later. What are your thoughts on ensuring the 30-minute expiration limit in this scenario as well? (Of course, the existing logic didn't account for this either.)

I think a change like the following might work, though I'm not entirely sure which approach is best:

```
final long boundedTimestamp = Math.max(nowMillis - MAX_TIMESTAMP_SHIFT_MILLIS, Math.min(**nowMillis**, timestamp));
```

renatsaf: You're right, and your version is the one I took — pushed in 6006e7e.

With the old upper clamp a `timestamp` five minutes ahead of the clock made the shift negative, so `lastAccessedNanos` landed five minutes in the future and a 30-minute session lived for 35. Clamping to `nowMillis` puts the shift in `[0, MAX_TIMESTAMP_SHIFT_MILLIS]`, so the monotonic access time only ever moves back and the session expires at most one timeout after the call. That also makes the `+ MAX_TIMESTAMP_SHIFT_MILLIS` bound unreachable, so I dropped it; the lower bound still keeps the nanosecond arithmetic in `isExpired` free of overflow.

Two notes on the semantics, in case you'd rather they went the other way:

- It's a behaviour change for callers of this public method, not only hardening: setting a future timestamp can no longer extend a session past its timeout. `setSessionTimeout` is the intended lever for that, and the servlet contract measures inactivity, so I think capping is right.
- `getTimestamp()` / `HttpSession#getLastAccessedTime()` still return the value as given — the clamp applies to expiry only. The same was already true for the lower bound, and I've said so in the javadoc rather than rewriting what the getter reports, since a setter that silently changes its argument would be more surprising.

Added `setTimestampInFutureDoesNotExtendBeyondTimeout` to `SessionExpirationTest`; it fails on the previous code and passes now.

carryel (approved):

PR #2315, opened 2026-09-14
Do not write GOAWAY to an already closed connection

### LINKED ISSUE #2314: `Unable to write GOAWAY.  Terminating session.` logged at end of HTTP/2 session
# How to reproduce

Please use [the reproducer from #2304](https://github.com/user-attachments/files/32052048/grizzly-2304-reproducer.zip), which is to be compiled using `mvn clean package`; alternatively use [the pre-built EAR](https://github.com/user-attachments/files/32052211/quipsy-ng-ear.zip).

Deploy it into Payara 7.2026.8 using `asadmin deploy --upload quipsy-ng-ear.ear`.

Enable HTTP/2 and TLS on `http-listener-1`.

Run `h2load -D 1 https://quipsy:8080/.well-known/quipsy`

Check Payara log file using `tail -n 25 <PAYARA_HOME>/glassfish/domains/domain1/logs/server.log`. *👉 Note that the problem only occurs when using `-D 1`, but will **not** occur when instead using `-n 1`!👈*

# Expected result

* No severe problems found in Payara log.

# Actual result

* Log files contains WARNING: `Unable to write GOAWAY.  Terminating session.` and a stack trace.

## Stack Trace

```
[2026-09-14T09:41:30.002+0200] [Payara 7.2026.8] [WARNUNG] [] [org.glassfish.grizzly.http2.Http2Session] [tid: _ThreadID=110 _ThreadName=http-thread-pool::http-listener-1(16)] [timeMillis: 1789371690002] [levelValue: 900] [[
  Unable to write GOAWAY.  Terminating session.
java.io.EOFException
        at org.glassfish.grizzly.nio.transport.TCPNIOTransport.read(TCPNIOTransport.java:597)
        at org.glassfish.grizzly.nio.transport.TCPNIOTransportFilter.handleRead(TCPNIOTransportFilter.java:59)
        at org.glassfish.grizzly.filterchain.TransportFilter.handleRead(TransportFilter.java:134)
        at org.glassfish.grizzly.ssl.SSLBaseFilter$SSLTransportFilterWrapper.handleRead(SSLBaseFilter.java:927)
        at org.glassfish.grizzly.filterchain.ExecutorResolver$9.execute(ExecutorResolver.java:89)
        at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeFilter(DefaultFilterChain.java:246)
        at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeChainPart(DefaultFilterChain.java:178)
        at org.glassfish.grizzly.filterchain.DefaultFilterChain.execute(DefaultFilterChain.java:118)
        at org.glassfish.grizzly.filterchain.DefaultFilterChain.process(DefaultFilterChain.java:96)
        at org.glassfish.grizzly.ProcessorExecutor.execute(ProcessorExecutor.java:51)
        at org.glassfish.grizzly.nio.transport.TCPNIOTransport.fireIOEvent(TCPNIOTransport.java:513)
        at org.glassfish.grizzly.strategies.AbstractIOStrategy.fireIOEvent(AbstractIOStrategy.java:82)
        at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.run0(WorkerThreadIOStrategy.java:83)
        at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy$WorkerThreadRunnable.run(WorkerThreadIOStrategy.java:101)
        at org.glassfish.grizzly.threadpool.AbstractThreadPool$Worker.doWork(AbstractThreadPool.java:540)
        at org.glassfish.grizzly.threadpool.AbstractThreadPool$Worker.run(AbstractThreadPool.java:520)
        at java.base/java.lang.Thread.run(Unknown Source)
]]
```

renatsaf:
Fixes #2314

## Cause

When the peer drops the TCP connection, `TCPNIOTransport.read()` turns the EOF into an `EOFException` and stores it as the connection's close reason:

```java
final IOException e = new EOFException();
tcpConnection.terminate0(null, new CloseReason(CloseType.REMOTELY, e));
throw e;
```

`Http2Session.ConnectionCloseListener` then sets `closeFlag` and calls `closedRemotely()` on every live stream. Each stream close cascades into `Http2Session.deregisterStream()`, which sees `isGoingAway() == true` and a stream count of zero, and calls `terminate(NO_ERROR, "Session closed")`.

The close listener sets `closeFlag` but leaves `goingAwayLastStreamId` at `Integer.MIN_VALUE`, so the "already in GOAWAY state" check in `setGoAwayLocally()` does not fire and a GOAWAY frame is built. It is written to a connection that has been dead since the EOF, so `AbstractNIOAsyncQueueWriter` fails the write with the connection's close reason:

```java
if (!nioConnection.isOpen()) {
    onWriteFailure(nioConnection, queueRecord, nioConnection.getCloseReason().getCause());
```

The completion handler logs that at `WARNING` with the stack trace of the `EOFException` — which is a *read*-path stack, not a write-path one, which is why the reported trace looks unrelated to GOAWAY. One such entry is produced per client that disconnects without a clean shutdown, which is what `h2load -D 1` does on every connection.

The second caller, `setGoAwayByPeer()`, answers a peer GOAWAY with its own GOAWAY and has the same hole.

## Change

`sendGoAwayAndClose()` now skips the write when the connection is no longer open and just terminates the session, logging at `FINE`. Both callers benefit.

The check cannot be exhaustive — the peer may close in between the check and the write — so `failed()` also logs at `FINE`, without the stack trace, when the connection turns out to be closed. A write failure on a connection that is still open keeps the `WARNING`.

## Tests

Two cases added to `Http2SessionTest`, along with the harness changes they need (a non-empty session filter chain, an `obtainProcessorState` stub, and log record capture):

- `testTerminateSkipsGoAwayOnClosedConnection` — no frame reaches the filter chain, the connection is closed, and nothing is logged at `INFO` or above. Fails on unmodified `Http2Session`.
- `testTerminateWritesGoAwayOnOpenConnection` — the GOAWAY is still written when the connection is live.

`mvn -pl modules/http2 -am test` is green (~470 tests). The harness itself used to emit the `Unable to write GOAWAY` warning during the test run; that is gone as well.


mkarg:
I am happy to confirm that `-D 1` does not reproduce the reported error anymore using this PR within Payara 7.2028.8. Thank you, @renatsaf! 🚀

OndroMih (approved):

modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2Session.java
```diff
@@ -579,32 +579,51 @@ void terminate(final ErrorCode errorCode, final String detail) {
...
-                private void close() {
-                    connection.closeSilently();
-                    outputSink.close();
-                }
+        if (!connection.isOpen()) {
+            // The peer is already gone, so there's nobody left to receive the
+            // GOAWAY. Writing it would only fail with the IOException the
+            // connection was closed with, which is not a problem worth reporting.
+            LOGGER.log(Level.FINE, "Connection is already closed.  Skipping GOAWAY and terminating session.");
+            frame.recycle();
```
carryel, line 592: It seems like it would be good to call `frame.recycle()` not only here but also when `closeConnectionAndOutputSink()` is executed as the `CompletionHandler` is called.

renatsaf: The frame is already recycled on that path. `Http2SessionOutputSink.writeDownStream(K, CompletionHandler, MessageCloner)` passes it to `Http2FrameCodec.serializeAndRecycle()`, which calls `frame.recycle()` right after `toBuffer()` and before the chain write. By the time the completion handler runs, the `GoAwayFrame` is back in the `ThreadCache`; the handler only reports the outcome of writing the serialized `Buffer`.

A second `recycle()` from the handler would put the same instance into the cache twice (`ObjectCacheElement.put()` does not check for duplicates), so two later `GoAwayFrame.create()` calls could return the same object. The handler may also run after the frame has already been taken from the cache again, and would then reset a frame that is in use.

The `frame.recycle()` in the early-return branch is needed only because that branch bypasses `serializeAndRecycle()`. Added a comment there to make that explicit in 2b28be8.

carryel (approved):

PR #2316, opened 2026-09-15
Allow %5C (encoded backslash) in request URIs per RFC 3986

### LINKED ISSUE #2028: grizzly-http silently fails when encoded backslashes are included in the URI
I don't have full steps to reproduce because the way our application is configured is somewhat complicated, but I do know that the problem lies in `HttpRequestURIDecoder::normalizeBytes` (and `normalizeChars`) - see [here](https://github.com/eclipse-ee4j/grizzly/blob/5d4371cfaa3dd6483e1422f812821658d78d5583/modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java#L525).

Essentially, since the URI is normalized _after_ the URI decoding happens, you have a case where an encoded backslash (`%5C`) is converted to a forward slash _or_ the `normalizeBytes` method returns `false` and then grizzly silently swallows an error (unless you have logging set to `FINE`) before returning a 500 response.

In fact, the whole idea of decoding before the path has been broken up into components seems completely backwards, since any encoded forward slashes (`%2F`) will be decoded and then subsequently treated as normal slashes, which they should not be. Unless, of course, you have encoded forward slashes disabled, which I understand was implemented to plug a security hole that I am _guessing_ was caused by decoding the URI before breaking up the path into its components.

lprimak:
`HttpRequestURIDecoder` percent-decodes the request URI before normalizing it, so `%5C` becomes `\` and `normalizeBytes`/`normalizeBuffer`/`normalizeChars` reject the request. The only alternative was the compile-time `ALLOW_BACKSLASH` constant, which rewrote `\` to `/` — turning path data into a delimiter, which RFC 3986 §2.4 says must not happen. In practice this means any JAX-RS path parameter containing a backslash (e.g. JSON with escaped quotes) gets a 400 from GlassFish/Payara, while the same app on standalone Jersey works because that path never calls the decoder.

**Changes**

- `ALLOW_BACKSLASH` is now read from the system property `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH`, alongside the existing `com.sun.enterprise.web.collapseAdjacentSlashes`.
- When enabled, a decoded backslash is kept as ordinary path data rather than rewritten to `/`. `checkNormalize()` is gated on the same flag so `convertToChars()` doesn't reject the URI one step later.
- A literal, unencoded `\` in the raw request URI is still rejected regardless of the flag — it isn't a valid URI character. The check runs on the raw URI because after decoding the two cases are indistinguishable. Only `%5C` is affected by this change.

**Default is unchanged (`false`).** A decoded `\` is a path separator to filesystem-backed handlers on Windows, and downstream guards such as Catalina's `/WEB-INF` check compare the path before that translation happens. Flipping the default would re-open that route for embedders that haven't hardened their guards, so this is opt-in, same as Tomcat's `allowBackslash`.

**Testing**

Verified against Payara 7 with a JAX-RS echo resource and `curl --path-as-is`:

| Request | Before | After (flag on) |
|---|---|---|
| `/echo/foo%5Cbar` | 400 | 200 `foo\bar` |
| `/echo/a%5C..%5Cb` | 400 | 200 `a\..\b` |
| `/echo/foo\bar` (literal) | 400 | 400 |
| `/echo/../WEB-INF/web.xml` | 404 | 404 |

Behavior with the flag off is byte-for-byte identical to before.

fixes #2028


mkarg:
@OndroMih Kindly asking for review! 🙏 I have waited for this fix so many years, and Lenny finally did it! ❤️

OndroMih:
Hi, @mkarg , I would like to but no time right now. I’m doing already a lot of opensource work for free, I’ll review this when I get to it.

Maybe some other @eclipse-ee4j/ee4j-glassfish-committers or @eclipse-ee4j/ee4j-grizzly-committers  woúd have time to review. The Grizzly project is actice, be patient and somebody will review.

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
```diff
@@ -328,9 +332,7 @@ public static boolean normalizeChars(final CharChunk uriCC) {
         // Check for null char
         for (pos = start; pos < end; pos++) {
             if (c[pos] == '\\') {
-                if (ALLOW_BACKSLASH) {
-                    c[pos] = '/';
-                } else {
+                if (!ALLOW_BACKSLASH) {
```
carryel: For the sake of consistency, couldn't we use a single `if` statement, like on line 287?

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
```diff
@@ -512,9 +514,7 @@ public static boolean normalizeBytes(final ByteChunk bc) {
                 return false;
             }
             if (bs[pos] == (byte) '\\') {
-                if (ALLOW_BACKSLASH) {
-                    bs[pos] = (byte) '/';
-                } else {
+                if (!ALLOW_BACKSLASH) {
```
carryel: For the sake of consistency, couldn't we use a single `if` statement, like on line 287?

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
```diff
@@ -614,9 +614,7 @@ public static boolean normalizeBuffer(final BufferChunk bc) {
                 return false;
             }
             if (b == (byte) '\\') {
-                if (ALLOW_BACKSLASH) {
-                    bs.put(pos, (byte) '/');
-                } else {
+                if (!ALLOW_BACKSLASH) {
```
carryel: For the sake of consistency, couldn't we use a single `if` statement, like on line 287?

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
```diff
@@ -44,7 +44,8 @@
  */
 public class HttpRequestURIDecoder {
 
-    protected static final boolean ALLOW_BACKSLASH = false;
+    // When true, a %5C that decoded to '\' is kept as path data (RFC 3986) instead of rejecting the URI.
+    protected static final boolean ALLOW_BACKSLASH = Boolean.parseBoolean(System.getProperty("com.sun.enterprise.web.allowBackslash", "false"));
```
carryel: Since this is a new property name anyway—and assuming it isn't already in use—it might be better to follow the convention used for other property names (such as `HttpRequestURIDecoder.class.getName()`) rather than using `com.sun.enterprise.web`.

See:
- https://github.com/eclipse-ee4j/glassfish-grizzly/blob/318799fc81540d7ab437059b493efe6820aca428/modules/http/src/main/java/org/glassfish/grizzly/http/util/Constants.java#L112
- https://github.com/eclipse-ee4j/glassfish-grizzly/blob/318799fc81540d7ab437059b493efe6820aca428/modules/http/src/main/java/org/glassfish/grizzly/http/util/UDecoder.java#L32-L33

carryel:
Additionally, it would be great to have relevant utility test cases as well, if possible!

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
lprimak: Done

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
lprimak: Done

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
lprimak: Done

modules/http/src/main/java/org/glassfish/grizzly/http/util/HttpRequestURIDecoder.java
lprimak: done

dmatej (approved):

lprimak:
Thank you David!

lprimak:
All comments have been addressed, and tests added. Thank you @carryel

lprimak:
Thank you @carryel and David!

